PR #8, opened 2016-02-12
Consider day of week of services

nkuehnel:
Also check if the given date is an active weekday. When checking for active services, the day of week was previously not yet considered.


landonreed:
Thanks @nkuehnel, this was incorporated via 73a4d4cf56469f0f8f20a19b4287b625f1be7628.

PR #22, opened 2016-06-21
Fixes wrong initialization of feedId

### LINKED ISSUE #21: FeedId is required now and is not properly initialized if it is missing in GTFS feed
I think that there was some merge problem or something in GTFSFeed since code around [line 127](https://github.com/conveyal/gtfs-lib/blob/master/src/main/java/com/conveyal/gtfs/GTFSFeed.java#L127) is strange:
```java
// maybe we should just point to the feed object itself instead of its ID, and null out its stoptimes map after loading
if (fid != null) {
    feedId = fid; //This isn't using file name feed id but provided feed id since fid != null
    LOG.info("Feed ID is undefined, pester maintainers to include a feed ID. Using file name {}.", feedId); // TODO log an error, ideally feeds should include a feedID
}
else if (feedId == null) { //feedId is almost never null since by default it is empty string
    feedId = new File(zip.getName()).getName().replaceAll("\\.zip$", "");
    LOG.info("Feed ID is undefined, pester maintainers to include a feed ID. Using file name {}.", feedId); // TODO log an error, ideally feeds should include a feedID
}
else {
    LOG.info("Feed ID is '{}'.", feedId);
}
```

Bigger problem is that I get duplicate FeedId error since if feed doesn't have ID by default it is empty string which != null so second gtfs file throws duplicateFeed ID error.

buma:
Fixes #21. Also no more DuplicateFeedId exceptions in R5.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/GTFSFeed.java
+++ b/src/main/java/com/conveyal/gtfs/GTFSFeed.java
@@ -125,11 +125,11 @@ public class GTFSFeed implements Cloneable, Closeable {
 
         new FeedInfo.Loader(this).loadTable(zip);
         // maybe we should just point to the feed object itself instead of its ID, and null out its stoptimes map after loading
-        if (fid != null) {
+        if (fid != null && !fid.isEmpty()) {
             feedId = fid;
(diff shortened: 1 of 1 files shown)


buma:
Already fixed by matt.

PR #24, opened 2016-10-05
Nysdot stats

landonreed:
Statistics and reports for NYSDOT.  More info on exact requirements [here](https://docs.google.com/document/d/1DJ5VBZxu1XYPMF-cvtFfjOkWdVUqla-06jvxqjD0-Fw/edit#).


src/main/java/com/conveyal/gtfs/GTFSFeed.java
```diff
@@ -707,6 +709,26 @@ public double getTripSpeed (String trip_id, boolean straightLine) {
...
 
+    /** Returns sorted map of stop_ids mapped to tuple keys of trip_id, stop_sequence (for use in feed.stop_times) */
+    public SortedSet<Fun.Tuple2<String, Fun.Tuple2>> getStopTimesForStop (String stop_id) {
+        SortedSet<Fun.Tuple2<String, Fun.Tuple2>> index = this.stopStopTimeSet
+                .subSet(new Fun.Tuple2<>(stop_id, null), new Fun.Tuple2(stop_id, Fun.HI));
+
+        return index;
+    }
+
+    public ZoneId getTimeZoneForStop (String stop_id) {
```
mattwigway: I think this function is named in confusing way; it implies that it's returning stop_timezone but it's returning agency_timezone.

src/main/java/com/conveyal/gtfs/stats/FeedStats.java
```diff
@@ -145,6 +154,40 @@ public LocalDate getEndDate() {
...
+        return this.pattern.getTotalRevenueTimeForTrips(getTripsForDate(date));
+    }
+
+    /** Get total revenue distance (in meters) for all trips on a given date. */
+    public double getTotalRevenueDistanceForTrips (LocalDate date) {
+        return this.pattern.getTotalRevenueDistanceForTrips(getTripsForDate(date));
+    }
+
+    /** in seconds */
+    public double getDailyAverageHeadway (LocalDate date, LocalTime from, LocalTime to) {
```
mattwigway: I question the value of this metric. It seems completely dependent on your bounds and how the GTFS is coded (e.g. are platforms coded separately...)

mattwigway: Also units are unclear.

src/main/java/com/conveyal/gtfs/stats/FeedStats.java
```diff
@@ -145,6 +154,40 @@ public LocalDate getEndDate() {
...
+
+        OptionalDouble avg =  feed.stops.values().stream()
+                .map(s -> this.stop.getAverageHeadwayForStop(s.stop_id, date, from, to))
+                .mapToDouble(headway -> headway)
+                .average();
+
+        return avg.getAsDouble();
+    }
+
+    public double getAverageTripSpeed (LocalTime from, LocalTime to) {
```
mattwigway: What are the units? Assuming furlongs per fortnight. Can we rename to getAverageTripSpeedKmh, or at least add some Javadoc to the method? (I generally prefer to put units right in the variables/function names that contain them so it's really clear, avoids Mars Climate Orbiter-type events).

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+
+    private GTFSFeed feed = null;
+    private FeedStats stats = null;
+
+    public PatternStats (GTFSFeed f, FeedStats fs) {
+        feed = f;
+        stats = fs;
+    }
+
+    public double getPatternSpeed (String pattern_id, LocalDate date, LocalTime from, LocalTime to) {
```
mattwigway: What are the units?

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+
+        List<Trip> trips = getTripsForDate(pattern_id, date);
+
+        return getAverageSpeedForTrips(trips, from, to);
+    }
+
+    public double getAverageSpeedForTrips (Collection<Trip> trips, LocalTime from, LocalTime to) {
```
mattwigway: Units?

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+    }
+
+    public LocalTime getStartTimeForTrips (Collection<Trip> trips) {
+        int earliestDeparture = Integer.MAX_VALUE;
+
+        for (Trip trip : trips) {
+            StopTime st;
+            try {
+                // use interpolated times in case our common stop is not a time point
+                st = StreamSupport.stream(feed.getInterpolatedStopTimesForTrip(trip.trip_id).spliterator(), false)
```
mattwigway: If we're using the first stop, why are we using interpolated times? It's illegal for the first stop to not have times specified.

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+                        .findFirst()
+                        .orElse(null);
+            } catch (GTFSFeed.FirstAndLastStopsDoNotHaveTimes e) {
+                return null;
+            }
+
+            // shouldn't actually encounter any departures after midnight, but skip any departures that do
+            if (st.departure_time > 86399) continue;
```
mattwigway: Believe this is actually legal but it's a trip that runs the next morning, so I'd say ignoring it is fine, but might clarify the comment.

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+    }
+
+    public LocalTime getEndTimeForTrips (Collection<Trip> trips) {
+        int latestArrival = Integer.MIN_VALUE;
+
+        for (Trip trip : trips) {
+            StopTime st;
+            try {
+                // use interpolated times in case our common stop is not a time point
+                st = StreamSupport.stream(feed.getInterpolatedStopTimesForTrip(trip.trip_id).spliterator(), false)
```
mattwigway: Same comment about interpolated times.

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+            if (st.arrival_time >= latestArrival) {
+                latestArrival = st.arrival_time;
+            }
+        }
+
+        // return end time as 2:00 am if last arrival occurs after midnight
+        return LocalTime.ofSecondOfDay(latestArrival % 86399);
+    }
+
+    /** Get total revenue time (in seconds) for set of trips. */
```
mattwigway: Yay! Units!

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+            }
+
+            first = StreamSupport.stream(interpolated, false)
+                    .findFirst()
+                    .orElse(null);
+
+            last = StreamSupport.stream(interpolated, false)
+                    .reduce((a, b) -> b)
+                    .orElse(null);
+            int time = last.departure_time - first.arrival_time;
```
mattwigway: Shouldn't this be `last.arrival_time - first.departure_time`? I wouldn't consider layover at first and last stops to be revenue time.

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+
+    /** Get total revenue time (in seconds) for set of trips. */
+    public int getTotalRevenueTimeForTrips (Collection<Trip> trips) {
+        TIntList times = new TIntArrayList();
+        for (Trip trip : trips) {
+            StopTime first;
+            StopTime last;
+            Spliterator<StopTime> interpolated = null;
+            try {
+                interpolated = feed.getInterpolatedStopTimesForTrip(trip.trip_id).spliterator();
```
mattwigway: Interpolated stop times are probably unnecessary here as well.

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+            int time = last.departure_time - first.arrival_time;
+
+            times.add(time);
+        }
+
+        return times.sum();
+    }
+
+    /** Get total revenue distance (in meters) for set of trips. */
+    public double getTotalRevenueDistanceForTrips (Collection<Trip> trips) {
```
mattwigway: Is there a difference between trip distance and revenue distance?

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+        TDoubleList distances = new TDoubleArrayList();
+        for (Trip trip : trips) {
+            distances.add(feed.getTripDistance(trip.trip_id, false));
+        }
+
+        return distances.sum();
+    }
+
+    public double getPatternDistance (String pattern_id) {
```
mattwigway: Units!

src/main/java/com/conveyal/gtfs/stats/PatternStats.java
```diff
@@ -1,7 +1,176 @@
...
+        Pattern pattern = feed.patterns.get(pattern_id);
+
+        return feed.getTripDistance(pattern.associatedTrips.iterator().next(), false);
+    }
+
+    public double getAverageStopSpacing (String pattern_id) {
```
mattwigway: Units!

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+    private RouteStats routeStats = null;
+
+    public StopStats (GTFSFeed f, FeedStats fs) {
+        feed = f;
+        stats = fs;
+        routeStats = stats.route;
+    }
+
+    public Set<Route> getRoutes (String stop_id) {
+        return feed.patterns.values().stream()
```
mattwigway: Is this actually a statistic?

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+
+    /** Get list of trips for specified date of service */
+    public List<Trip> getTripsForDate (String stop_id, LocalDate date) {
+        List<String> tripIds = stats.getTripsForDate(date).stream()
+                .map(trip -> trip.trip_id)
+                .collect(Collectors.toList());
+
+        return feed.getStopTimesForStop(stop_id).stream()
+                .map(t -> feed.trips.get(t.b.a)) // map to trip ids
+                .filter(t -> tripIds.contains(t.trip_id)) // filter by trip_id list for date
```
mattwigway: Possibility of duplicates here due to loop routes, recommend adding .distinct()

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+                .collect(Collectors.toList());
+    }
+
+    public int getAverageHeadwayForStop (String stop_id, LocalDate date, LocalTime from, LocalTime to) {
+        List<Trip> tripsForStop = feed.getStopTimesForStop(stop_id).stream()
```
mattwigway: Seems like we used this same code before, let's just call getTripsForDate here (unless you were intending to use from/to here and it's been omitted). I like this form better, we're not looking up all the trips on a date to use a filter.

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+                .map(trip -> trip.trip_id)
+                .collect(Collectors.toList());
+
+        return feed.getStopTimesForStop(stop_id).stream()
+                .map(t -> feed.trips.get(t.b.a)) // map to trip ids
+                .filter(t -> tripIds.contains(t.trip_id)) // filter by trip_id list for date
+                .collect(Collectors.toList());
+    }
+
+    public int getAverageHeadwayForStop (String stop_id, LocalDate date, LocalTime from, LocalTime to) {
```
mattwigway: Units!

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+        List<Trip> tripsForStop = feed.getStopTimesForStop(stop_id).stream()
+                .map(t -> feed.trips.get(t.b.a))
+                .filter(trip -> feed.services.get(trip.service_id).activeOn(date))
+                .collect(Collectors.toList());
+
+        return getStopHeadwayForTrips(stop_id, tripsForStop, from, to);
+    }
+
+    /** Get the route headway for a given service date at a stop over a time window, in seconds */
+    public Map<String, Integer> getRouteHeadwaysForStop (String stop_id, LocalDate date, LocalTime from, LocalTime to) {
```
mattwigway: Does this need to accomodate direction_ids? we're going to have biased results otherwise when we compare headways of (generally) bus service with separate stops per direction, and rail/busway service (which generally has one stop for both directions).

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+    public int getStopHeadwayForTrips (String stop_id, List<Trip> trips, LocalTime from, LocalTime to) {
+        TIntList timesAtStop = new TIntArrayList();
+
+        for (Trip trip : trips) {
+            StopTime st;
+            try {
+                // use interpolated times in case our common stop is not a time point
+                st = StreamSupport.stream(feed.getInterpolatedStopTimesForTrip(trip.trip_id).spliterator(), false)
+                        .filter(candidate -> candidate.stop_id.equals(stop_id))
+                        .findFirst()
```
mattwigway: This is looking at the first stop not the requested stop, which is going to mess up when you have multiple terminals.

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+
+            if (delta > 60) deltas.add(delta);
+        }
+
+        if (deltas.isEmpty()) return -1;
+
+        return deltas.sum() / deltas.size();
+    }
+
+    public int getHeadwayForStopByRoute (String stop_id, String route_id, LocalDate date, LocalTime from, LocalTime to) {
```
mattwigway: Units!

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+                        }
+                        times.add(st);
+                        routeStopTimeMap.put(route.route_id, times);
+                    }
+                });
+        Map<Fun.Tuple2<String, String>, TIntList> waitTimesByRoute = new HashMap<>();
+        Map<Fun.Tuple2<String, String>, Set<Fun.Tuple2<StopTime, StopTime>>> missedTransfers = new HashMap<>();
+        // iterate over every entry in route -> list of stopTimes map
+        for (Map.Entry<String, List<StopTime>> entry : routeStopTimeMap.entrySet()) {
+            final int MISSED_TRANSFER_THRESHOLD = 60 * 10;
```
mattwigway: Let's not bury this so far down in the code. Also, ten minutes seems arbitrary.

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+                                int waitTime = compareTime.departure_time - currentTime.arrival_time;
+
+                                // if non-negative and shortest, save for later
+                                if (waitTime >= 0 && waitTime < shortestWait){
+                                    shortestWait = waitTime;
+                                }
+                                // otherwise, check for missed near-transfer opportunities
+                                else {
+                                    if (waitTime < 0 && waitTime * -1 <= MISSED_TRANSFER_THRESHOLD) {
+                                        Fun.Tuple2<StopTime, StopTime> missedTransfer = new Fun.Tuple2(compareTime, currentTime);
```
mattwigway: can we just use int[2] here?

src/main/java/com/conveyal/gtfs/stats/StopStats.java
```diff
@@ -1,7 +1,239 @@
...
+                        for (StopTime compareTime : compareTimes) {
+                            if (compareTime.departure_time > 0) {
+                                int waitTime = compareTime.departure_time - currentTime.arrival_time;
+
+                                // if non-negative and shortest, save for later
+                                if (waitTime >= 0 && waitTime < shortestWait){
+                                    shortestWait = waitTime;
+                                }
+                                // otherwise, check for missed near-transfer opportunities
+                                else {
```
mattwigway: Couldn't the shortest wait also be a missed transfer, if all the times are missed transfers.

src/main/java/com/conveyal/gtfs/stats/model/MissedTransfer.java
```diff
@@ -0,0 +1,14 @@
+package com.conveyal.gtfs.stats.model;
+
+import java.util.Set;
+
+/**
+ * Created by landon on 10/4/16.
+ */
+public class MissedTransfer {
```
mattwigway: Is this used?

src/main/java/com/conveyal/gtfs/stats/model/StopStatistic.java
```diff
@@ -0,0 +1,25 @@
...
+import com.conveyal.gtfs.stats.StopStats;
+
+import java.time.LocalDate;
+import java.time.LocalTime;
+
+/**
+ * Created by landon on 10/4/16.
+ */
+public class StopStatistic {
+    public String stop_id;
```
mattwigway: Do we need a separate class or can we just use stopstats and mark things transient/@JsonIgnore?

src/main/java/com/conveyal/gtfs/stats/model/TransferPerformanceSummary.java
```diff
@@ -0,0 +1,31 @@
...
+        this.fromRoute = fromRoute;
+        this.toRoute = toRoute;
+        bestCase = minWaitTime;
+        worstCase = maxWaitTime;
+        typicalCase = avgWaitTime;
+        missedOpportunities = missedTransfers;
+    }
+
+    public String toString () {
+        return String.format("From routes %s to %s, the best case transfer time is %d seconds, worst case is %d seconds, and typical case is %d seconds. %d missed near-transfer opportunities.", fromRoute, toRoute, bestCase, worstCase, typicalCase, missedOpportunities.size());
```
mattwigway: It's not typical, it's average. The median would be typical #pedantry.

src/main/java/com/conveyal/gtfs/stats/FeedStats.java
mattwigway: http://humantransit.org/2010/08/the-perils-of-succeeding-on-average.html

src/main/java/com/conveyal/gtfs/GTFSFeed.java
```diff
@@ -707,6 +709,26 @@ public double getTripSpeed (String trip_id, boolean straightLine) {
         return distance / time; // meters per second
     }
 
+    /** Returns sorted map of stop_ids mapped to tuple keys of trip_id, stop_sequence (for use in feed.stop_times) */
+    public SortedSet<Fun.Tuple2<String, Fun.Tuple2>> getStopTimesForStop (String stop_id) {
```
landonreed: Maybe.  I had also considered that, but it seemed that it might be helpful to have the tuple with trip ID.  But now that I'm thinking about it again... you're right (trip ID and sequence are easy enough to pull out from StopTime objects themselves).

src/main/java/com/conveyal/gtfs/stats/StopStats.java
landonreed: No, definitely not a statistic.  Should we move this helper to GTFSFeed or do you think it's not worth keeping?

src/main/java/com/conveyal/gtfs/stats/StopStats.java
mattwigway: Yeah probably, although I wonder if there's a way to specify this without using an arbitrary parameter like this.

src/main/java/com/conveyal/gtfs/stats/StopStats.java
landonreed: Yea, perhaps it should.

src/main/java/com/conveyal/gtfs/stats/StopStats.java
landonreed: is that enough info?  seems like it might be useful to have at least trip ID associated with the times.

src/main/java/com/conveyal/gtfs/stats/model/MissedTransfer.java
landonreed: No, that was added unintentionally.

src/main/java/com/conveyal/gtfs/stats/model/StopStatistic.java
landonreed: Maybe we could just use JsonIgnore.

landonreed:
@abryd @trevorgerhardt I believe I made this PR after the fact because this code had been pushed to `master`.  However, I wanted @mattwigway to review, so he suggested I make the two branches by checking out `master` before and after the commits in question.

We used this to handle the code review, but at this point I believe this PR should be closed.

PR #28, opened 2016-11-22
fix(build): fix slf4j import/exclusion, and remove duplicate trove4j dep

### LINKED ISSUE #26: slf4j binding warning
As this project is a library the [logging dependencies](https://github.com/conveyal/gtfs-lib/blob/master/pom.xml#L128-L132) should be set to 'test' or 'runtime' scope to avoid binding the wrong slf4j binding and this warning: http://www.slf4j.org/codes.html#multiple_bindings

As a workaround one can use:
```xml
        <dependency>
            <groupId>com.conveyal</groupId>
            <artifactId>gtfs-lib</artifactId>
            <version>1.2.0</version>

            <exclusions>
                <exclusion> 
                    <groupId>org.slf4j</groupId>
                    <artifactId>slf4j-simple</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

landonreed:
fixes #26

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -214,18 +214,6 @@
   </repositories>
 
   <dependencies>
-    <dependency>
-      <groupId>com.conveyal</groupId>
-      <artifactId>gtfs-lib</artifactId>
-      <version>${project.version}</version>
-
-      <exclusions>
-        <exclusion>
-          <groupId>org.slf4j</groupId>
-          <artifactId>slf4j-simple</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
     <!-- CSV parser -->
     <dependency>
       <groupId>net.sourceforge.javacsv</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -86,6 +86,18 @@
     </repository>
   </repositories>
   <dependencies>
+    <dependency>
+      <groupId>com.conveyal</groupId>
+      <artifactId>gtfs-lib</artifactId>
```
mattwigway: Why are we adding gtfs-lib as a dependency of itself?

landonreed:
How's this @mattwigway?

landonreed:
How's this @mattwigway?

mattwigway:
This looks good to me. @karussell this look good to you?

karussell:
Yes, thanks! This should work.

Issue #35, opened 2017-03-08
Feed errors missing because line # set to zero


landonreed:
A number of `GTFSError` subclasses have call `super("trips", 0, "trip_id")`, where the middle argument line number is set to `0` because there is no line number to reference (the error is caught after loading).  This is causing the errors to be dropped because of `GTFSError::compareTo`.

Here is a list of all the classes that need to be fixed:
![image](https://cloud.githubusercontent.com/assets/2370911/23718976/86429058-0407-11e7-9217-1a70a828375d.png)

`RouteNameError` also needs to be fixed.

PR #36, opened 2017-03-10
Error line fixes + enhanced CLI + validation json output

landonreed:
This PR addresses #35 (missing errors because of incomplete `compareTo`) as well as a few other things:

- enhanced CLI usage (as well as shaded jar) with options for running and saving validation
- JSON serializers (for use in saving validation and eventually to be imported by conveyal/datatools-server)
- fix for ReversedTripValidator uninentional NPEs


src/main/java/com/conveyal/gtfs/error/DuplicateStopError.java
```diff
@@ -11,9 +11,9 @@
     private final String message;
     public final DuplicateStops duplicateStop;
 
-    public DuplicateStopError(String message, DuplicateStops duplicateStop) {
-        super("stop", 0, "stop_lat,stop_lon");
-        this.message = message;
+    public DuplicateStopError(DuplicateStops duplicateStop) {
+        super("stop", duplicateStop.stop1.sourceFileLine, "stop_lat,stop_lon", duplicateStop.getStop1Id());
```
mattwigway: I think we want to pick which stop to use for the source line and affected entity based on which has a higher sourceFileLine. The second stop in the file is the duplicate.

src/main/java/com/conveyal/gtfs/error/DuplicateTripError.java
```diff
@@ -12,21 +12,19 @@
...
-    public final String tripId;
     public final String duplicateTripId;
     public final String tripKey;
     public final String routeId;
 
-    public DuplicateTripError(String tripId, String duplicateTripId, String tripKey, String routeId) {
-        super("trips", 0, "trip_id");
-        this.tripId = tripId;
+    public DuplicateTripError(Trip trip, String duplicateTripId, String tripKey) {
+        super("trips", trip.sourceFileLine, "trip_id", trip.trip_id);
```
mattwigway: Above comment about choosing which trip is the duplicate applies, but will have to be done where this is called.

src/main/java/com/conveyal/gtfs/error/DuplicateTripError.java
```diff
@@ -12,21 +12,19 @@
...
         this.duplicateTripId = duplicateTripId;
         this.tripKey = tripKey;
-        this.routeId = routeId;
+        this.routeId = trip.route_id;
 
     }
 
     @Override public String getMessage() {
-        return "Trip Ids " + duplicateTripId + " & " + tripId + " are duplicates (" + tripKey + ")";
+        return String.format("Trip Ids %s & %s (route %s) are duplicates (%s)", duplicateTripId, affectedEntityId, routeId, tripKey);
```
mattwigway: Might make sense to use the human-readable name of the pattern here, rather than the trip key.

src/main/java/com/conveyal/gtfs/error/OverlappingTripsInBlockError.java
```diff
@@ -12,16 +12,16 @@
     public static final long serialVersionUID = 1L;
 
     public final String[] tripIds;
-    public final String affectedEntityId;
     public final Priority priority = Priority.HIGH;
+    public final String routeId;
 
-    public OverlappingTripsInBlockError(long line, String field, String affectedEntityId, Route route, String[] tripIds, Priority priority) {
-        super("trips", line, field);
+    public OverlappingTripsInBlockError(long line, String field, String affectedEntityId, String routeId, String[] tripIds) {
```
mattwigway: Is the affected entity the route, or the block?

src/main/java/com/conveyal/gtfs/error/TimeZoneError.java
```diff
@@ -11,11 +11,9 @@
     public static final long serialVersionUID = 1L;
 
     public final String message;
-    public final String affectedEntityId;
     public TimeZoneError(String tableName, long line, String field, String affectedEntityId, String message) {
```
mattwigway: This at least needs javadoc describing what affectedEntityId is (route or stop)

src/main/java/com/conveyal/gtfs/error/UnusedStopError.java
```diff
@@ -9,13 +9,11 @@
 public class UnusedStopError extends GTFSError implements Serializable {
     public static final long serialVersionUID = 1L;
 
-    public final String affectedEntityId;
     public final Priority priority;
     public final Stop stop;
 
     public UnusedStopError(String affectedEntityId, long line, Stop stop) {
```
mattwigway: Isn't affectedEntityId just stop.stop_id here?

src/main/java/com/conveyal/gtfs/model/Entity.java
```diff
@@ -92,7 +93,7 @@ private String getFieldCheckRequired(String column, boolean required) throws IOE
             String str = reader.get(column);
             if (str == null) {
                 if (!missingRequiredColumns.contains(column)) {
-                    feed.errors.add(new MissingColumnError(tableName, column));
+                    feed.errors.add(new MissingColumnError(tableName, row, column));
```
mattwigway: Won't this just always be row 0?

src/main/java/com/conveyal/gtfs/util/json/JsonManager.java
```diff
@@ -0,0 +1,104 @@
+package com.conveyal.gtfs.util.json;
```
mattwigway: This is probably an overcomplicated way to handle JSON serialization but it's fine for now.

src/main/java/com/conveyal/gtfs/util/json/JsonViews.java
```diff
@@ -0,0 +1,19 @@
+package com.conveyal.gtfs.util.json;
+
+/**
+ * Defines all of the JSON views.
+ * This is basically just a list.
+ * @author mattwigway
+ *
+ */
+public class JsonViews {
```
mattwigway: We probably don't need to use views at all in gtfs-lib.

src/main/java/com/conveyal/gtfs/validator/MissingStopCoordinatesValidator.java
```diff
@@ -16,12 +16,11 @@ public boolean validate(GTFSFeed feed, boolean repair) {
         boolean isValid = true;
         for (Stop stop : feed.stops.values()) {
             Coordinate stopCoord = new Coordinate(stop.stop_lat, stop.stop_lon);
-
-            ProjectedCoordinate projectedStopCoord = null;
             try {
-                projectedStopCoord = GeoUtils.convertLatLonToEuclidean(stopCoord);
+                // attempt to project coordinate
+                GeoUtils.convertLatLonToEuclidean(stopCoord);
```
mattwigway: We shouldn't be using this function anywhere anymore; it attempts to project points to an "appropriate" UTM zone. Also, we're reversing lat and lon above (I think it get reversed again inside this function).

src/main/java/com/conveyal/gtfs/validator/ReversedTripsValidator.java
```diff
@@ -61,64 +63,64 @@ public boolean validate(GTFSFeed feed, boolean repair) {
...
+            Geometry lastStopGeom;
+            Coordinate firstShapeCoord;
+            Coordinate lastShapeCoord;
+
+            // if coordinate creation fails here, add trip_id to missing shapes list
             try {
                 firstStopCoord = new Coordinate(feed.stops.get(firstStop.stop_id).stop_lat, feed.stops.get(firstStop.stop_id).stop_lon);
                 lastStopCoord = new Coordinate(feed.stops.get(lastStop.stop_id).stop_lat, feed.stops.get(lastStop.stop_id).stop_lon);
 
                 firstStopGeom = geometryFactory.createPoint(GeoUtils.convertLatLonToEuclidean(firstStopCoord));
```
mattwigway: This is dangerous if the first and last stops don't land in the same UTM zone.

src/main/java/com/conveyal/gtfs/validator/ReversedTripsValidator.java
```diff
@@ -36,21 +39,20 @@ public boolean validate(GTFSFeed feed, boolean repair, Double distanceMultiplier
     @Override
     public boolean validate(GTFSFeed feed, boolean repair) {
         boolean isValid = true;
-        ValidationResult result = new ValidationResult();
         int errorLimit = 5000;
```
mattwigway: What is this for?

mattwigway: Ah, just to avoid spewing too many errors. I think we shoudn't do this, errors should be filtered at display time but we shouldn't have hard limits on how many are generated.

src/main/java/com/conveyal/gtfs/validator/TripTimesValidator.java
```diff
@@ -22,13 +22,12 @@ public boolean validate(GTFSFeed feed, boolean repair) {
         // Trips should have at least one hop (at least two stops)
         // Speed should not be infinite (check distance, time separately)
         boolean isValid = true;
-        ValidationResult result = new ValidationResult();
         int errorLimit = 2000;
```
mattwigway: Same comment about error limit.

src/main/java/com/conveyal/gtfs/error/OverlappingTripsInBlockError.java
landonreed: Right, I'm thinking that it should be the `block_id`.  I'll change that param name.

src/main/java/com/conveyal/gtfs/error/UnusedStopError.java
landonreed: Yea, I missed that somehow.

src/main/java/com/conveyal/gtfs/validator/ReversedTripsValidator.java
landonreed: I added `errorLimit` after a discussion with @abyrd and you I think.  The idea was that if we're catching up to some large `n` of errors, then additional errors may not be valuable to have because there is probably something devastatingly wrong with the feed and it will rather just slow down loading of the feed (and make the validation report excessively large and burdensome to download).

We can change this if we think it's not actually a burden to churn through all of the errors, but just wanted to get the original purpose out there.

src/main/java/com/conveyal/gtfs/error/DuplicateStopError.java
landonreed: @mattwigway Is there a good shorthand way to do this in the `super` call (without a long, ugly ternary statement)?

src/main/java/com/conveyal/gtfs/error/DuplicateTripError.java
landonreed: This mimics the message provided in the old validator lib, fyi.  Maybe there was a reason it was defined that way originally?  I'm not really sure.

src/main/java/com/conveyal/gtfs/validator/MissingStopCoordinatesValidator.java
landonreed: @mattwigway what do you propose instead?

src/main/java/com/conveyal/gtfs/error/TimeZoneError.java
landonreed: Do you mean: agency or stop?

src/main/java/com/conveyal/gtfs/error/DuplicateTripError.java
mattwigway: I'm just saying for display purposes we should use the human-readable name. Also, we could just check if the trips have the same pattern ID, rather than looping over all the stops.

src/main/java/com/conveyal/gtfs/error/TimeZoneError.java
mattwigway: yup

src/main/java/com/conveyal/gtfs/validator/MissingStopCoordinatesValidator.java
mattwigway: I think here we can just check if stop_lat and stop_lon are NaN, right? We might need to adjust the loader so missing values are NaN.

src/main/java/com/conveyal/gtfs/validator/ReversedTripsValidator.java
mattwigway: Fair enough, I rest my case.

src/main/java/com/conveyal/gtfs/error/DuplicateStopError.java
mattwigway: Make functions getOriginalStop and getDuplicatedStop in the DuplicateStops class?

src/main/java/com/conveyal/gtfs/validator/MissingStopCoordinatesValidator.java
landonreed: @mattwigway, Actually, I think we can just get rid of this validator, no?  The existence of valid lat/lon values is already validated on load.

src/main/java/com/conveyal/gtfs/error/MissingColumnError.java
```diff
@@ -6,8 +6,8 @@
 public class MissingColumnError extends GTFSError implements Serializable {
     public static final long serialVersionUID = 1L;
 
-    public MissingColumnError(String file, long line, String field) {
-        super(file, line, field);
+    public MissingColumnError(String file, String field) {
+        super(file, 0, field);
```
mattwigway: Should this actually be line 1? Maybe it's not important.

PR #37, opened 2017-03-16
Start and end transit dates are saved

buma:
I added feedStart and feedEnd date to gtfs-lib. With that added I can update R5 so that during building/opening of graph start/end of transit data can be shown. Since at least during debugging I tend to forgot for what dates I have transit times.

A little problem I noticed is with Washington data since CUE has data from 2000 to 2099 which are min/max dates of all feeds and report of start/end feed dates is useless. I'm not sure what to do about that. One option is to just save start/end feed dates if they are around 2-3 years of current years.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/GTFSFeed.java
+++ b/src/main/java/com/conveyal/gtfs/GTFSFeed.java
@@ -135,6 +135,10 @@ public class GTFSFeed implements Cloneable, Closeable {
     /* A place to store an event bus that is passed through constructor. */
     public transient EventBus eventBus;
 
+    public LocalDate feedStartDate = LocalDate.MAX;
+
+    public LocalDate feedEndDate = LocalDate.MIN;
+
     /**
      * The order in which we load the tables is important for two reasons.
      * 1. We must load feed_info first so we know the feed ID before loading any other entities. This could be relaxed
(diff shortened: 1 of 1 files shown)


abyrd:
I'm going to do this as part of the trips-per-date index, so I'll probably close this PR as soon as it's implemented elsewhere.

trevorgerhardt:
@abyrd where are we heading with this? This will be useful for https://github.com/conveyal/analysis-ui/issues/478

abyrd:
@trevorgerhardt  gtfs-lib will create a table of the quantity of stop_times, trips, and "revenue hours"  per day of defined service.

abyrd:
As an update, version 3.1 does produce the "revenue hours" table and is a step in the direction of accurately guessing when the feed is valid.

abyrd:
@landonreed I think the only proper way to do this is to calculate "revenue hours" per day and look for patterns/weekly changes, treating any feed_info start/end dates as hints about when its validity starts and ends. I've seen too many cases where only part of the service was running at the beginning or end of the feed.

This particular patch fixes MapDB-based GTFSFeed, which I don't think we're going to be spending as much time updating or maintaining as the new RDBMS-backed GTFS class (unless it's to merge the two).

So I'd say we should close this, I'd like to avoid this method.

PR #42, opened 2017-04-13
Add GTFSCache constructor with cacheSize

landonreed:


src/main/java/com/conveyal/gtfs/GTFSCache.java
```diff
@@ -80,6 +59,51 @@ public GTFSCache(String bucket, String bucketFolder, File cacheDir) {
...
+        this.cacheDir = cacheDir;
+        createCache(cacheSize);
+    }
+
+    /**
+     * Create cache of GTFSFeed objects with variable size.
+     * of
+     * @param cacheSize maximum number of feeds to be stored in cache
+     */
+    private void createCache(int cacheSize) {
```
mattwigway: Rather than having a separate function to do this, I think it probably makes more sense to have a chained constructors, so you'd have:

GTFSCacche(bucket, cacheDir) which calls GTFSCache(bucket, bucketFolder, cacheDir) which calls GTFSCache(bucket, folder, cacheDir, cacheSize).

src/main/java/com/conveyal/gtfs/GTFSCache.java
```diff
@@ -248,6 +274,7 @@ private GTFSFeed retrieveFeed (String originalId) {
                 throw new RuntimeException(e);
             }
         } else {
+            LOG.warn("Feed {} found neither locally nor on S3", originalId);
```
mattwigway: If it's not found on S3, won't we throw a runtimeexception above?

landonreed: Well, this is reached when no file is found whatsoever, so I though it would be most descriptive to say `neither local nor s3`.

mattwigway: Is it reached though if it's not on S3? Won't we have already thrown an exception?

src/main/java/com/conveyal/gtfs/GTFSCache.java
```diff
@@ -96,14 +74,18 @@ private void createCache(int cacheSize) {
...
+                .build(new CacheLoader<String, GTFSFeed>() {
+                    @Override
+                    public GTFSFeed load(String s) throws Exception {
+                        return retrieveFeed(s);
+                    }
+                });
+    }
+
+    public long getCurrentCacheSize() {
+        return this.cache.size();
```
mattwigway: Did you intend to have this return the max size? I think that this will return the number of items currently in the cache (which will be 0 when the server first starts up).

landonreed: No, I meant for it to return the current size (in case it's needed at some point?).  Max size was meant to be grabbed from `GTFSCache.cacheSize`.

src/main/java/com/conveyal/gtfs/GTFSCache.java
landonreed: Ah, yea I guess you're right.

src/main/java/com/conveyal/gtfs/GTFSCache.java
mattwigway: :+1:

PR #45, opened 2017-05-10
Fix run command in README

barbeau:
* "-validate" must come after the JAR file name, otherwise you get an error "Error: A fatal exception has occurred. Program will exit. Unrecognized option: -validate"
* Specify exact version in file name example, with a note that it should be changed to match /target directory


landonreed:
LGTM, thanks @barbeau!

PR #51, opened 2017-05-10
fix(Route): Fix agency NPE issue

### LINKED ISSUE #40: NullPointerException without error description if agency id is empty
The current switzerland gtfs has a few problematic lines with an empty agency_id, e.g.:

"66-140-Y-j17-1","","140-Y","","0"

Which breaks the code in Route.java with a NPE at [the last line](https://github.com/conveyal/gtfs-lib/blob/master/src/main/java/com/conveyal/gtfs/model/Route.java#L57):

```java
          Agency agency = (Agency)this.getRefField("agency_id", false, this.feed.agency);
            if(agency == null && this.feed.agency.size() == 1) {
                agency = (Agency)this.feed.agency.values().iterator().next();
            }

            r.agency_id = agency.agency_id;
```

I would prefer throwing a proper exception instead or similar:

```java
          Agency agency = (Agency)this.getRefField("agency_id", false, this.feed.agency);
            if(agency == null) {
                if(this.feed.agency.size() == 1)
                  agency = (Agency)this.feed.agency.values().iterator().next();
                else
                  throw new RuntimeException("Cannot find agency for route "+r.route_id+". Routes contain invalid data at row " + (this.row+1));
            }
```

landonreed:
This fix is for feeds that have no agency records. When the agency_id for a route has a missing
reference, it is assigned the first agency if there is only one agency. If there are no agencies
present, the field is left null and a NoOperatorInFeedError is stored.

Fixes #40

### DIFF
--- a/src/main/java/com/conveyal/gtfs/error/NoOperatorInFeedError.java
+++ b/src/main/java/com/conveyal/gtfs/error/NoOperatorInFeedError.java
@@ -1,0 +1,18 @@
+package com.conveyal.gtfs.error;
+
+import com.conveyal.gtfs.validator.model.Priority;
+
+/**
+ * Created by landon on 5/2/17.
+ */
+public class NoOperatorInFeedError extends GTFSError {
+    public final Priority priority = Priority.HIGH;
+
+    public NoOperatorInFeedError() {
+        super("agency", 0, "agency_id");
+    }
+
+    @Override public String getMessage() {
+        return String.format("No operator listed in feed (must have at least one).");
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/com/conveyal/gtfs/error/NoOperatorInFeedError.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+public class NoOperatorInFeedError extends GTFSError {
+    public final Priority priority = Priority.HIGH;
+
+    public NoOperatorInFeedError() {
+        super("agency", 0, "agency_id");
+    }
+
+    @Override public String getMessage() {
+        return String.format("No operator listed in feed (must have at least one).");
```
mattwigway: Should this say "operator" or, should it "agency" to align with GTFS terms?

landonreed: I'll change the name of the file, too.

PR #52, opened 2017-05-11
Use System.nanoTime() to measure elapsed time

barbeau:
* See "Conclusion" of https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks, which says:

>If you are interested in measuring/calculating elapsed time, then always use System.nanoTime().


landonreed:
Thanks @barbeau. LGTM

PR #59, opened 2017-08-17
Fare Attributes : missing transfers means "unlimited"

zishanbilal:
Missing transfers means "unlimited" in this case use value grater-than 2 (for now i supplies/mapped it to 3), rather than 0. As 0 already has some meaning, so it cause bug/mislead the lib user.


zishanbilal:
@abyrd Please review the code or assign some review.

abyrd:
@zishanbilal we should discuss your workflow for integrating changes/bugfixes from dependency libraries. Because even when we merge a PR like this into the master branch, of course it will get built and deployed, but as a snapshot release, so your local branch of R5 or anything else using gtfs-lib would need to depend on that snapshot release. Maybe that's what you're doing... It would just be good for us to understand how you're handling this.

PR #64, opened 2017-09-19
Add tests to project

evansiroky:
### Bug fix

Make sure `stats.FeedStats.getAverageWeekdayRevenueTime` doesn't throw a divide by 0 exception when given a new Feed.

#### stats.FeedStatsTest

Adds a unit test to make sure the `getAverageWeekdayRevenueTime` doesn't throw a divide by 0 exception when given a new Feed.

#### util.UtilTest

I added a super simple first test case to kick off this project by testing the human function.

#### GeometriesTest

Makes tests for the Netherlands geometry utils.  I added tests here because it substantially increases the coverage of the library since the file it's testing is so big.

#### GTFSFeedTest

Attempt to do a roundtrip of loading a GTFS zip file and then writing another zip file.

#### GTFSMainTest

Has a test that can make sure the main method can load a GTFS zip.

#### GTFSTest 

This test suite is intended to be a main entry point from which to do high-level integration tests of all kinds of possible GTFS files.  Currently there is a method called `runSimpleIntegrationTest` which simply runs the loading and vaidation of a GTFS zip file.  If any part fails, the test fails, so it is very simple at this point.


.travis.yml
```diff
@@ -16,6 +16,13 @@ before_install: |
     gpg --import --batch maven-artifact-signing-key.asc
   fi
 
+before_script:
+  - psql -U postgres -c 'CREATE DATABASE gtfs_lib_test;'
```
abyrd: Should we use the same programmatic create and drop that you used in osm-lib? It seems more portable (would work on a local build for example, as long as you had postgres running).

.travis.yml
```diff
@@ -27,6 +34,9 @@ script: |
     mvn clean verify --settings maven-settings.xml -B -V -Dgpg.skip
   fi
 
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
abyrd: [cringe]

pom.xml
```diff
@@ -243,6 +249,11 @@
     	<version>1.13</version>
     </dependency>
     <dependency>
+      <groupId>org.hamcrest</groupId>
```
abyrd: Please add XML comments explaining what the dependencies are.

src/test/java/com/conveyal/gtfs/GTFSFeedTest.java
```diff
@@ -0,0 +1,27 @@
...
+    public void canDoRoundtripLoadAndWriteToZipFile() {
+        String outZip = getResourceFileName("fake-agency-output.zip");
+        File file = new File(outZip);
+        file.delete();
+        try {
+            GTFSFeed feed = new GTFSFeed();
+            feed.fromFile(getResourceFileName("fake-agency.zip"));
+            feed.toFile(outZip);
+            assertThat(file.exists(), equalTo(true));
+            // TODO: assert that rows of data were written
```
abyrd: Ideally we'd then read it back in and do some comparison. There should already be tests that do this for the MapDB backed code.

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Tests whether or not a super simple 2-stop, 1-route, 1-trip, valid gtfs can be loaded
+     */
+    @Test
+    public void canLoadSimpleAgency() {
+        runSimpleIntegrationTest("fake-agency.zip");
+    }
+
+    /**
+     * Run GTFS.main with a certain zip file.
```
abyrd: This javadoc does not explain what the method is for.  We should avoid trivial Javadoc. The purpose appears to be loading and validating the specified feed.

.travis.yml
evansiroky: I removed this

.travis.yml
evansiroky: I asked a question about this on stackoverflow: 
 https://stackoverflow.com/questions/46308469/security-concerns-about-running-scripts-on-travis-ci

src/test/java/com/conveyal/gtfs/GTFSFeedTest.java
evansiroky: I added some tests for this, but they are failing because data isn't being written.  I'm not sure if I'm using the feed instance correctly here.

.travis.yml
abyrd: Good idea to ask for input on Stack Overflow. It seemed to me that the curl-pipe-bash method was born out of a mixture of laziness and naivete, but would be interested in getting alternate perspectives. The comments on this one sum up the usual reaction I've seen to such things ("horrible", "inadmissible" etc.) https://gist.github.com/btm/6700524

abyrd:
Other than the concerns about `curl | bash` (for which I don't think there's a much better solution) this looks pretty good. Should we merge and continue to add more tests later?

PR #66, opened 2017-09-20
Add RDBMS support (Version 3.0)

abyrd:
This branch adds the ability to load GTFS data into a relational database instead of a MapDB file. The loading code targets a wide range of RDBMS and therefore attempts to use very generic SQL statements, avoiding special features. The exception is the Postgres LOAD command, which makes loading large feeds much faster but requires a strange alternate code path where everything is converted back into untyped text and piped over to the database server.

Some capabilities of RDMBS are not used fully. For example, many validators just iterate over all the entities in a particular category with the validation logic happening in Java, rather than doing the validation logic in SQL queries. This is intentional to preserve backward compatibility with MapDB and plain maps, in case we want to use it someday.

Each GTFS feed loaded creates a separate schema in the database, and the txt files in the GTFS feed become tables in that schema. This allows each loaded feed to have different columns, reflecting any extensions present in the feed. Validation results are stored in another table inside the schema.

There is one shared "feeds" table that keeps track of all the feeds that have been loaded and associates each one of them with a unique schema name.

When merged, these changes will become version 3.0 of gtfs-lib. We will keep the MapDB support because it's useful for scratch files and modifications (e.g. loading a feed into R5, filtering and cropping a feed for use in an analysis project).


pom.xml
```diff
@@ -265,6 +265,11 @@
     	<version>2.4</version>
     </dependency>
     <dependency>
+      <groupId>commons-dbutils</groupId>
+      <artifactId>commons-dbutils</artifactId>
+      <version>1.6</version>
+    </dependency>
```
evansiroky: Please add XML comments explaining what the dependencies are.

pom.xml
```diff
@@ -287,14 +292,44 @@
...
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
+      <version>1.4.194</version>
+    </dependency>
+    <dependency>
+      <groupId>com.j256</groupId>
+      <artifactId>ormlite</artifactId>
+      <version>2.9</version>
+    </dependency>
```
evansiroky: Please add XML comments explaining what the dependencies are.

src/main/java/com/conveyal/gtfs/GTFS.java
```diff
@@ -0,0 +1,202 @@
...
+                .hasArg().argName("password").desc("database password").build());
+        options.addOption(Option.builder().longOpt("graphql")
+                .desc("start a GraphQL API on the given port").optionalArg(true).build());
+        options.addOption(Option.builder().longOpt("json").desc("optionally store in result.json").build());
+        return options;
+    }
+
+    private static void printHelp(Options options) {
+        final String HELP = String.join("\n",
+                "java -cp gtfs-lib-shaded.jar com.conveyal.gtfs.GTFS [options]",
```
evansiroky: Please add more descriptions of what the options are.

src/main/java/com/conveyal/gtfs/loader/DateField.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public DateField (String name, Requirement requirement) {
+        super(name, requirement);
+    }
+
+    private int validate (String string) {
+        if (string.length() != 8) {
+            throw new StorageException("Date field should be exactly 8 characters long.");
+        }
+        int year = Integer.parseInt(string.substring(0, 4));
```
evansiroky: It would probably be better to use a date parser for this.  That way we'd also find problems such as trying to parse the date February 30th.

src/main/java/com/conveyal/gtfs/model/Entity.java
```diff
@@ -49,11 +49,27 @@
...
     /* The feed from which this entity was loaded. */
     transient GTFSFeed feed;
 
+    /**
+     * @return a key that according to the GTFS spec should uniquely identify this entity, either alone or together
+     *          with a sequence number. For example stop_times and shapes have no single field that uniquely
+     *          identifies a row, and the stop_sequence or shape_pt_sequence must also be considered.
+     */
+    public String getId () {
+        return null;
```
evansiroky: Looks like a WIP, but just leaving a comment to make sure this gets implemented or removed.

src/main/java/com/conveyal/gtfs/model/Entity.java
```diff
@@ -49,11 +49,27 @@
...
+    }
+
+    /**
+     * @return the integer second element of a compound key, for those elements that require one.
+     */
+    public Integer getSequenceNumber () {
+        return null;
```
evansiroky: Looks like a WIP, but just leaving a comment to make sure this gets implemented or removed.

src/main/java/com/conveyal/gtfs/storage/Backend.java
```diff
@@ -0,0 +1,93 @@
...
+            connection.commit();
+        } catch (Exception ex) {
+            throw new StorageException(ex);
+        }
+    }
+
+    public static void main(String args[]) throws Exception {
+//        Backend backend = new Backend();
+//        backend.setup();
+        GTFSFeed feed = GTFSFeed.fromFile("/Users/abyrd/geodata/nl/NL-OPENOV-20170322-gtfs.zip");
```
evansiroky: If these main methods are just examples, we should instead be creating tests instead.

src/main/java/com/conveyal/gtfs/storage/BooleanAsciiGrid.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public boolean getValueForCoords (double x, double y) {
+        int xCell = (int)((x - xllcorner) / cellsize);
+        int yCell = (int)((y - yllcorner) / cellsize);
+        int index = (nrows - yCell - 1) * ncols + xCell; // Vertical flip
+        if (index < 0 || index > grid.length()) return false;
+        return grid.get(index);
+    }
+
+    // TODO make this into a test
```
evansiroky: 💯

src/main/java/com/conveyal/gtfs/storage/ObjectRelationalMapping.java
```diff
@@ -0,0 +1,163 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.Statement;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Created by abyrd on 2017-03-27
```
evansiroky: We should not reinvent the wheel by creating an ORM.  👎

src/main/resources/com/conveyal/gtfs/storage/sql_statements.sql
```diff
@@ -0,0 +1,32 @@
+-- create_feeds_table
```
evansiroky: What is the purpose of this file?

pom.xml
abyrd: Done, and some dependencies removed entirely.

src/main/java/com/conveyal/gtfs/GTFS.java
abyrd: The command line parser library will include the descriptions of the options. Here's the current output:
```
usage: java -cp gtfs-lib-shaded.jar com.conveyal.gtfs.GTFS [options]

 -d,--database <url>        JDBC URL for the database. Defaults to
                            jdbc:postgresql://localhost/gtfs
    --graphql               start a GraphQL API on the given port
 -h,--help                  print this message
    --json                  optionally store in result.json
    --load <file>           load GTFS data from the given file
 -p,--password <password>   database password
 -u,--user <username>       database username
    --validate <feed>       validate the specified feed. defaults to the
                            feed loaded with the --load option
```

src/main/java/com/conveyal/gtfs/loader/DateField.java
abyrd: I changed this to use a LocalDate parser and added a test to check that it works right.

src/main/java/com/conveyal/gtfs/model/Entity.java
abyrd: This is in fact already used. The method is overridden on subclasses. I have added a comment to this effect.

src/main/java/com/conveyal/gtfs/model/Entity.java
abyrd: As above, already used. I added a comment explaining it's overridden on subclasses. For example, on stop times:

```
@Override
    public Integer getSequenceNumber() {
        return stop_sequence; // Compound key of StopTime is (trip_id, stop_sequence)
    }

```

src/main/java/com/conveyal/gtfs/storage/Backend.java
abyrd: Actually this is an early experiment and should be deleted.

src/main/java/com/conveyal/gtfs/storage/BooleanAsciiGrid.java
abyrd: Made it into a test.

src/main/java/com/conveyal/gtfs/storage/ObjectRelationalMapping.java
abyrd: This was an early experiment and has been removed.

src/main/resources/com/conveyal/gtfs/storage/sql_statements.sql
abyrd: I was originally loading pre-written SQL statements from a file before I started generating them programmatically. I'm deleting this file.

abyrd:
I'm going to merge this and cut release 3.0 which does not have truly complete RDBMS support but allows us to pull all our other software forward to the existing MapDB implementation that's in 3.0, which should remain completely unchanged in the future.

PR #77, opened 2017-10-20
Fix #75 sql loader row counts

### LINKED ISSUE #75: Off by two error in sql loader row counts
When loading a feed, there appears to be an off by two error for row counts.

For example:
- agency.txt with one row yields `-1` rowCount
- routes.text with 11 routes yields `-9` rowCount

[example feed (GCT)](https://github.com/conveyal/gtfs-lib/files/1363503/GCT-20171006T114955-04-0b3b23bf-3dd7-4e32-b173-ae13910c0f4a.zip)

aguiraf:
Fix sql loader row counts


landonreed:
Thanks @aguiraf, this looks good to me and seems to fix the issue for my local dev version.  @abyrd, any thoughts here?  If not, I'll go ahead and merge.

PR #82, opened 2017-11-14
Numerous validation error fixes

landonreed:
## A few changes
- Use symbolic constant `INT_MISSING` for missing ints in GTFS entities
- implement `getId` and `getSequenceNumber` abstract methods for Entity classes
- handle non-UTF-8 character encodings


abyrd:
@landonreed I made some changes to the UTF-8 check. Instead of comparing strings we should use the constant objects representing charsets.

PR #87, opened 2017-11-21
Integrate graphql API directly into gtfs-lib

abyrd:
It used to be in a separate project called gtfs-api.
We just put the Java graphql api definition in this project, which takes string queries and returns Object responses.
Whatever project includes gtfs-lib will need to provide an HTTP layer on top of this using some web server or framework.


abyrd:
@landonreed feel free to merge dev to master and cut a point release (as outlined in Javastyle) if you want to integrate into datatools-server.

PR #91, opened 2017-12-08
Set locale to US on humanize test

### LINKED ISSUE #89: The test for humanizing numbers failed
`UtilTest.canHumanize()` test failed in `Locale.FRANCE`. 
`Locale` must be set to `Locale.US`: 
Add `java.uti;.Locale.setDefault(java.util.Locale.US);` on top of `UtilTest.canHumanize()` function.``

maukoquiroga:
Fixes #89


abyrd:
Thanks @maukoquiroga. I guess the test only demonstrates that this works in one locale, but it's not a very important test. I think it was added just because we've aiming at eventually having full code coverage.

PR #92, opened 2017-12-11
Wheelchair boarding can be 2

l-vincent-l:
It's specified here: https://developers.google.com/transit/gtfs/reference/#stopstxt

PR #95, opened 2018-01-04
Fix table.java according to GTFS specification

l-vincent-l:
I double checked the Table.java against the GTFS specification to be sure everything was ok.
I found a few mistakes.


landonreed:
@l-vincent-l, same comment as I made in #92 about the conflicts.

PR #99, opened 2018-02-13
[WIP] Fix table.java according to GTFS specification

l-vincent-l:
I need to recheck that.


landonreed:
LGTM, @l-vincent-l.  Thanks!

landonreed:
Oh, I'm seeing that this is marked as WIP.  Are there more changes you need to add, @l-vincent-l?

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -129,15 +130,15 @@ public Table (String name, Class<? extends Entity> entityClass, Requirement requ
         new StringField("route_short_name",  OPTIONAL), // one of short or long must be provided
         new StringField("route_long_name",  OPTIONAL),
         new StringField("route_desc",  OPTIONAL),
-        // FIXME: Should the route type max value be equivalent to GTFS spec's max?
-        new IntegerField("route_type", REQUIRED, 999),
+        new IntegerField("route_type", REQUIRED, 7),
```
landonreed: @abyrd, I think you wrote the original version of this code. I think this PR is good to go, but wanted to see if there was some reason to keep the max value for `routes#route_type` as 999.

abyrd: No, there's no justification for the number 999. The GTFS spec only includes whole numbers up to 7, but there was a never-really-adopted extension that would use TPEG codes: https://groups.google.com/forum/#!msg/gtfs-changes/keT5rTPS7Y0/71uMz2l6ke0J

That would allow values up to 1700-ish. I think 1800 is a good upper limit.

PR #98, opened 2018-02-09
Mapdb feed to sql

landonreed:
This PR adds a converter from the MapDB-based GTFSFeed to SQL database.  Among the changes are:

- Adds `ScheduleException` entity type (for the GTFS editor).
- Adds `setParameters` abstract method for `Entity` and its subtypes that sets the parameters on a SQL prepared statement.
- Fixes SQL insertion of ints with the `Entity.INT_MISSING` value.
- Adds `autoCommit` flag to `JdbcTableWriter` to allow for multiple writes in a single transaction.

PR #100, opened 2018-02-15
A few fixes for the GTFS Editor

landonreed:
### Stops for pattern and stops for route fields added to the GraphQL schema.
- This allows for fetching GTFS stop entities that are served by a pattern or a route nested underneath one of these entities.
### Adds blank snapshot functionality.
- This allows for the creation of a "blank" snapshot, i.e., the full set of tables required for GTFS editing with no rows.


landonreed:
@abyrd, merging for expediency, but flagging for review.  This one is not as big as #98, but both this and #98 would benefit from your review.

PR #101, opened 2018-02-18
Re-implement findPatterns method in MapDB GTFSFeed using storage-agnostic PatternFinder

landonreed:
This fixes gtfs-lib for usage in r5.  The primary thing broken during the SQL load/edit enhancements was `GTFSFeed#findPatterns`, which was never re-implemented using the new `PatternFinder` class. This temporarily patches this so that r5 does not break when using the latest `gtfs-lib`.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/GTFSFeed.java
+++ b/src/main/java/com/conveyal/gtfs/GTFSFeed.java
@@ -49,8 +49,6 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import java.util.zip.ZipOutputStream;
 
-import static com.conveyal.gtfs.util.Util.human;
-
 /**
  * All entities must be from a single feed namespace.
  * Composed of several GTFSTables.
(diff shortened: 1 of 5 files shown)


src/main/java/com/conveyal/gtfs/GTFSFeed.java
```diff
@@ -397,6 +395,23 @@ public Shape getShape (String shape_id) {
         return shape.shape_dist_traveled.length > 0 ? shape : null;
     }
 
+    /**
+     * MapDB-based implementation to find patterns.
+     *
+     * FIXME: Remove and make pattern finding happen during validation?p
```
abyrd: I believe this comment sums up the reason the original method was removed. I wanted to share the pattern finder between the two implementations (MapDB and RDBMS), apply the same validation process to both kinds of storage, and produce Patterns in the same way in both cases, during validation. This prevents us from iterating over every stopTime twice, since we're already iterating over all of them in validation. This comment should probably be expanded to explain the intended path of development.

landonreed: OK, I will expand the comment to include this.  It's worth noting, however, that the stop times for MapDB have already been placed in a map where it is easy to retrieve them.

PR #102, opened 2018-02-20
Filter approved routes when exporting from GTFS editor

landonreed:
This PR adds the GTFS editor feature that filters out un-approved routes when exporting a snapshot to GTFS.  This is achieved by embedding filtering select statements into the `COPY TO` SQL.  These filters select only those routes with `status = 2` (i.e., route is approved) and joins this selection of routes to related tables (specifically, trips, stop_times, and shapes).

It also addresses a bug where duplicate row headers or headers named `id` cause a table load failure.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
+++ b/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
@@ -19,9 +19,18 @@ import java.io.FileOutputStream;
 import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.net.URI;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
@@ -40,7 +49,7 @@ public class JdbcGtfsExporter {
     private ZipOutputStream zipOutputStream;
     // The reference feed ID (namespace) to copy.
     private final String feedIdToExport;
-    private boolean postgresText;
+    private List<String> emptyTableList = new ArrayList<>();
 
     public JdbcGtfsExporter(String feedId, String outFile, DataSource dataSource, boolean fromEditor) {
         this.feedIdToExport = feedId;
@@ -68,16 +77,18 @@ public class JdbcGtfsExporter {
             // If we create a schema or table on one connection, then access it in a separate connection, we have no
             // guarantee that it exists when the accessing statement is executed.
             connection = dataSource.getConnection();
-            // Use the Postgres text load format if we're connected to that DBMS.
-            postgresText = connection.getMetaData().getDatabaseProductName().equals("PostgreSQL");
-            // Construct where clause for routes for use in filtering GTFS data if exporting from the editor.
-            String routesWhereClause = String.format(" where %s.%s.status = 2", feedIdToExport, Table.ROUTES.name);
+            if (!connection.getMetaData().getDatabaseProductName().equals("PostgreSQL")) {
+                // This code path currently requires the Postgres text "copy to" format.
+                throw new RuntimeException("Export from SQL database not implemented for non-PostgreSQL databases.");
+            }
+            // Construct where clause for routes table to filter only "approved" routes (and entities related to routes)
+            // if exporting a feed/schema that represents an editor snapshot.
+            String whereRouteIsApproved = String.format("where %s.%s.status = 2", feedIdToExport, Table.ROUTES.name);
             // Export each table in turn (by placing entry in zip output stream).
-            // FIXME: NO non-fatal exception errors are being captured during copy operations.
             result.agency = export(Table.AGENCY);
             result.calendar = export(Table.CALENDAR);
             if (fromEditor) {
-                // Export schedule exceptions in place of calendar dates if exporting from the GTFS Editor.
+                // Export schedule exceptions in place of calendar dates if exporting a feed/schema that represents an editor snapshot.
                 GTFSFeed feed = new GTFSFeed();
                 // FIXME: The below table readers should probably just share a connection with the exporter.
                 JDBCTableReader<ScheduleException> exceptionsReader =
@@ -87,6 +98,7 @@ public class JdbcGtfsExporter {
                         new JDBCTableReader(Table.CALENDAR, dataSource, feedIdToExport + ".",
                                 EntityPopulator.CALENDAR);
                 Iterable<Calendar> calendars = calendarsReader.getAll();
+                int calendarDateCount = 0;
                 for (Calendar cal : calendars) {
                     LOG.info("Iterating over calendar {}", cal.service_id);
                     Service service = new Service(cal.service_id);
@@ -114,12 +126,17 @@ public class JdbcGtfsExporter {
                                 throw new IllegalArgumentException("Duplicate schedule exceptions on " + date.toString());
 
                             service.calendar_dates.put(date, calendarDate);
+                            calendarDateCount += 1;
                         }
                     }
                     feed.services.put(cal.service_id, service);
                 }
-                LOG.info("Writing calendar dates from schedule exceptions");
-                new CalendarDate.Writer(feed).writeTable(zipOutputStream);
+                if (calendarDateCount == 0) {
+                    LOG.info("No calendar dates found. Skipping table.");
+                } else {
+                    LOG.info("Writing {} calendar dates from schedule exceptions", calendarDateCount);
+                    new CalendarDate.Writer(feed).writeTable(zipOutputStream);
+                }
             } else {
                 // Otherwise, simply export the calendar dates as they were loaded in.
                 result.calendarDates = export(Table.CALENDAR_DATES);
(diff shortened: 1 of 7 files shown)


src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -67,8 +68,10 @@ public FeedLoadResult exportTables() {
             // If we create a schema or table on one connection, then access it in a separate connection, we have no
             // guarantee that it exists when the accessing statement is executed.
             connection = dataSource.getConnection();
-            // Include the dot separator in the table prefix.
-            // This allows everything to work even when there's no prefix.
+            // Use the Postgres text load format if we're connected to that DBMS.
+            postgresText = connection.getMetaData().getDatabaseProductName().equals("PostgreSQL");
+            // Construct where clause for routes for use in filtering GTFS data if exporting from the editor.
```
abyrd: The information in the comment can be gleaned from the variable name `routesWhereClause` and the general purpose of `where` clauses. The more important thing to document is why you would want to filter, especially since the SQL string contains a magic number 2. Something like "Construct an SQL where clause that retains only routes that have been flagged for export to public GTFS". Is there a symbolic constant somewhere for that number 2?

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -124,26 +127,70 @@ public FeedLoadResult exportTables() {
...
-            result.frequencies = export(Table.FREQUENCIES);
-            // FIXME: only write "approved" routes using COPY TO with results of select query
-            result.routes = export(Table.ROUTES);
+            // Only write frequencies for "approved" routes using COPY TO with results of select query
+            String frequencySelectSql = null;
+            if (fromEditor) {
+                // Generate filter SQL for trips if exporting from the editor.
+                // The filter clause for frequencies requires two joins to reach the routes table and a where filter on
+                // route status.
+                frequencySelectSql = Table.FREQUENCIES.generateSelectSql(feedIdToExport, Requirement.OPTIONAL) +
```
abyrd: Rather than concatenating the strings with the + operator it might make more sense from a reader's point of view to pass them all to join() with spaces as the separator.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -124,26 +127,70 @@ public FeedLoadResult exportTables() {
...
+                        routesWhereClause;
+            }
+            result.shapes = export(Table.SHAPES, shapeSelectSql);
             result.stops = export(Table.STOPS);
-            // FIXME: only write stop times for "approved" routes using COPY TO with results of select query
-            result.stopTimes = export(Table.STOP_TIMES);
+            // Only write stop times for "approved" routes using COPY TO with results of select query
+            String stopTimesSelectSql = null;
+            if (fromEditor) {
+                // Generate filter SQL for trips if exporting from the editor.
```
abyrd: What does it mean to export "from the editor"? I am guessing this means to export a schema of tables that constitute an editor snapshot, not necessarily from the editor UI. Slightly more specific terminology should be used throughout.

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -31,6 +30,7 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(Table.class);
     private static final Requirement[] requirementsForEditorFields = new Requirement[]{EDITOR, REQUIRED, OPTIONAL};
+    private static final Requirement[] requirementsForSpecFields = new Requirement[]{REQUIRED, OPTIONAL};
```
abyrd: It's not clear what an "editor field" or "spec field" is. Please add documentation.

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -302,13 +304,27 @@ private Table withParentTable(Table parentTable) {
         return this;
     }
 
-    public List<Field> fieldsForEditor () {
+    public List<Field> editorFields() {
```
abyrd: See comments on specFields() below.

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -400,10 +426,48 @@ public String generateUpdateSql (String namespace, int id) {
     }
 
     /**
-     * Generate select all SQL string.
+     * Generate select all SQL string. The minimum requirement parameter is used to determine which fields ought to be
+     * included in the select statement. For example, if "OPTIONAL" is passed in, both optional and required fields
+     * are included in the select. If "EDITOR" is the minimum requirement, editor, optional, and required fields will
+     * all be included.
+     */
+    public String generateSelectSql (String namespace, Requirement minimumRequirement) {
```
abyrd: Via this method I think I understand what requirementsForSpecFields and requirementsForEditorFields mean. It appears that these and editorFields(), requiredFields(), and specFields() are fragmented implementations of an unstated underlying idea: there is a total order on the set of requirement enum values, (i.e. that the <= operator applies to all pairs of elements in that set), as well as an operation to extract the subset of elements less than or equal to a supplied element. If so, you have introduced the idea that the `Requirement` enum values are "levels" like log levels or Maven build cycle targets. If that's the case it should be spelled out in the comments, and reflected in the data structures and operators (methods) defined on those data types.

Concretely, the enum values should be placed in an ordered list or assigned order constants, which allow a generic filtering predicate that returns `true` for elements "up to and including" a supplied level.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -380,9 +386,16 @@ private int loadInternal (Table table) throws Exception {
             // The first field holds the line number of the CSV file. Prepared statement parameters are one-based.
             if (postgresText) transformedStrings[0] = Integer.toString(lineNumber);
             else insertStatement.setInt(1, lineNumber);
+
+            // Remove the values from the csv row for any bad fields (i.e., duplicate headers or named "ID").
+            // CSV values needs to be a LinkedList to support removal operation (Arrays.asList is read-only).
+            List<String> values = new LinkedList<>(Arrays.asList(csvReader.getValues()));
```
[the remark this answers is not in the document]
landonreed: @abyrd, I considered doing this, but got tripped up because I failed to consider that I could just maintain a separate column index for setting the prepared statement parameters (it must account for those null fields).  There's an easy way to do that though.  Instead of using `f` as the index, I'll just maintain a separate one.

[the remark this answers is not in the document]
landonreed: I still need some version of the `rawFields` array though because I need to create some other array with the clean set of fields to construct the Table definition.

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -302,13 +304,27 @@ private Table withParentTable(Table parentTable) {
...
+    public List<Field> editorFields() {
         // Filter out fields not used in editor (i.e., extension fields).
         return Arrays.stream(fields)
                 .filter(field -> Arrays.asList(requirementsForEditorFields).indexOf(field.requirement) != -1)
                 .collect(Collectors.toList());
     }
 
+    public List<Field> requiredFields () {
+        // Filter out fields not used in editor (i.e., extension fields).
+        return Arrays.stream(fields)
```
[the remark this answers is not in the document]
landonreed: Hmm, I don't feel too strongly about this either way, but I do agree that your syntax is a little easier to read.  I'll update.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
landonreed: Good point.  There is not a symbolic constant for the number 2.  I'm not sure where it would go.  What do you think?

src/main/java/com/conveyal/gtfs/loader/Table.java
landonreed: This will be removed in favor of the for loop approach you advocated.

src/main/java/com/conveyal/gtfs/loader/Table.java
landonreed: @abyrd, yes, generally that is my impression of how these requirements might ought to operate.  However, I'm not certain how all of the requirement types might be hierarchically arranged.  For now, I think we should capture your commentary on this in a `TODO` and have further conversation about where the other types (e.g., `PROPRIETARY`, `EXTENSION`, `UNKOWN`) might fit in here.

abyrd: Honestly I don't think there's a clear order to the requirement types. From the existing methods I got the impression you had an implicit ordering in mind, but if not then all the better, let's keep using clearly defined sets of requirement fields for each situation. In that case I just think those should be defined as EnumSets or some other appropriate abstraction instead of the current approach.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
abyrd: Also then you don't have to strategically inject spaces at the beginning and end of the SQL fragments generated by your methods.

src/main/java/com/conveyal/gtfs/loader/Table.java
abyrd: Just to clarify: a key part of that for loop approach was the inlined boolean expression instead of referencing an array of allowed types. In this case I think just inlining the few boolean tests is more readable.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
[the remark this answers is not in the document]
abyrd: I didn't think of this before. So you need both something to tell you which fields to skip and something to tell you which ones are present. Another idea: add an `int csvColumn` field to the `Field` class, make and iterate over a list of only the fields you want to load (indexes within that list are the prepared statement parameters), and draw the column number from the new field. But I'd lean toward iterating over the list of Fields with nulls in it, filtering out the nulls at the moment it's used in the Table definition.

Issue #97, opened 2018-02-06
Move to JTS 1.15.0


karussell:
JTS is now under the umbrella of locationtech and has a new release: https://github.com/locationtech/jts/

landonreed:
@abyrd, do you see any issues with updating the JTS version?

abyrd:
No, I wouldn't expect any problems with a minor version bump. But it would need to be tested before we use it in production.

willcohen:
This project uses geotools, which still has some dependencies on JTS 1.14. It seems like the main use of geotools here is [transforming points between projections](https://github.com/conveyal/gtfs-lib/blob/15ae095f9018ad4f4a7316f4a1eaf789ff498c2c/src/main/java/com/conveyal/gtfs/validator/service/GeoUtils.java#L73). Would you consider switching to something like [proj4j](https://github.com/Proj4J/proj4j) for projection, which would remove the dependency on geotools, not be LGPL, and not lock the JTS version?

abyrd:
I'm surveying the code for uses of Geotools. In GTFSFeed, GeodeticCalculator.orthodromicDistance is used in getInterpolatedStopTimesForTrip(). GeoUtils.getDistance() is used when measuring the length of a trip in meters. There is a ProjectedCoordinate class that doesn't seem to be used at all. There is a ReversedTripValidator that has not yet been reimplemented for the new validation framework, mostly  because it contains no documentation explaining what it's supposed to do. 

I think that's all the direct and indirect uses. All of them are just distance calculations, for which we have a standard, tolerably accurate method that doesn't involve rigorous projections.

abyrd:
Thanks for the summary @willcohen. 

Our business is based on providing the public sector with software that uses transparent reproducible methods, so we have no reverse engineering concerns. We do like to see widespread reuse of open data though, and know that many businesses are wary of LGPL (for reasons I've never been able to fully understand). 

I don't think this library exposes or returns any JTS objects via its intended public interface, so there should not be any issue with object incompatibility, where calling code that imports gtfs-lib as a dependency would be expecting JTS 1.15 objects. So it seems like the only real concern is gtfs-lib pulling in some unnecessary transitive dependencies. Can you confirm that?

Anyway, comments in our own code imply that we didn't really want to pull in JTS to do these distance calculations. We could certainly do it with another library or our usual equirectangular method. That's why I did the above survey of call sites to Geotools/JTS - I agree we should eliminate this dependency.

[...]

PR #103, opened 2018-02-22
Remove dependency on geotools and bump JTS to 1.16

willcohen:
Re #97, assuming the geotools dependency should be removed, this pull request replaces .getOrthodromicDistance in GTFSFeed with Util.fastDistance, modifies ReversedTripValidator to do distance comparisons using Util.fastDistance, removes the reprojection code from GeoUtils, takes out ProjectedCoordinate, and removes geotools.


src/main/java/com/conveyal/gtfs/validator/service/GeoUtils.java
```diff
@@ -1,185 +1,15 @@
...
-    //    spec[2] = "+ellps=clrk66";
-    //    spec[3] = "+units=m";
-    //    spec[4] = "+datum=NAD83";
-    //    spec[5] = "+no_defs";
-    //    Projection projection = ProjectionFactory.fromPROJ4Specification(spec);
-    //    return projection;
-  }
-
   /*
    * Taken from OneBusAway's UTMLibrary class
```
abyrd: comment is now incorrect

willcohen:
After bringing this PR up to date with master, a few tests are now failing, as can be seen in travis. I'm having a hard time diagnosing where the tests broke since HEAD is unable to complete `mvn test` or `mvn clean verify --settings maven-settings.xml -B -V -Dgpg.skip` on my local machine, yielding the error `Tests in error: 
  canLoadSimpleAgency(com.conveyal.gtfs.GTFSTest): org.postgresql.util.PSQLException: FATAL: database "null" does not exist`, which I can reproduce on a clean clone of the repository both at HEAD and at v4.0.0. Do I need to be generating additional test data before trying to run builds now?

landonreed:
@willcohen, I believe tests added by @evansiroky now require Postgres to be running to perform the test suite.  Does your system have Postgres installed and running?  If not this could be the reason `mvn test` is not completing.

willcohen:
A clarification question -- are you saying that GTFSFeed, in its entirety, can be fully removed? I started going through this a little while ago but it seemed like that there was a lot of code to be pulled.

abyrd:
@willcohen just chiming in to confirm what @landonreed said - we can't remove GTFSFeed. It's an almost completely separate legacy implementation / code path used by a lot of our software. It should eventually be merged into the newer GTFS class such that it can store feeds in either a RDBMS or MapDB (or just in memory, with memory being much more bountiful these days than when I first started working on this).

willcohen:
Great, thanks! So it sounds like the only classes that need to be actually removed, then, are the ones that are basically fully JTS-dependent, like Shape/ShapeMap/ShapePoint/Geometries. I've got a couple of other deadlines cooking right now but have gotten a start on this. It's true that there's not that many direct references to JTS in the library, but there's a fair number more references to these other classes across everything, so it does seem like removing JTS will touch a substantial portion of the library, even if all the MapDB functionality stays included. I'm, clearly, much less familiar with the inner workings of this library than you two, so this is a slower process for me. If you're eager to get this merged and wrapped up ASAP to get other breaking changes ready for another release, this larger dead code removal may be something I'd need some help with.

abyrd:
See comment on #97 - I've been working on this and think we need to keep JTS as a dependency of gtfs-lib. So let's just bump the JTS version to 1.16 and only remove Geotools.

willcohen:
Will bump today and we can get this resolved!

willcohen:
@abyrd let me know if there's anything else you need on this PR. A little uncertain of what additional minimization of JTS you're still looking for.

abyrd:
@willcohen I've just been reviewing this and it looks good to me. The minimization of JTS we were referring to was avoiding leaking JTS objects outside gtfs-lib to avoid disrupting software depending on gtfs-lib. But I now see that Shape and Pattern objects defined within gtfs-lib, which are certainly intended to be used outside the library, contain references to JTS geometries.

So we just need to accept that this change will require all systems depending on gtfs-lib to migrate to Locationtech JTS, which we should probably do anyway. 

@landonreed let me know if you have any other comments on this (e.g. how it might affect data-tools) before we merge it into dev.

landonreed:
@abyrd, the serialization issue looks to be with the `FeedStats#getMergedBuffers` field (which is contained within `ValidationResult` and serialized in the code snippet below). I don't know that anyone is actually using this (definitely no one within Conveyal is). Should we just remove this getMergedBuffers field?

https://github.com/conveyal/gtfs-lib/blob/42c11701297bb702e16bedc13e5ac0a53b37a07a/src/main/java/com/conveyal/gtfs/GTFSMain.java#L42-L45

abyrd:
@landonreed getMergedBuffers() looks like something intended for displaying the feed extent on a map, or identifying geographic overlap with other feeds. I suppose it could be useful in the future. On the other hand that's potentially a lot of information to save if we don't really use it for anything. I would like to see why serialization fails and possibly fix it.

I just re-checked to be sure I understand the implications on other projects. This version of JTS (1.16) is still the same one used by the latest version of Geotools (20.1) which we're migrating to in other projects, so it should be a good move.

abyrd:
Here is the error I see when saving validation info to JSON: `Exception in thread "main" com.fasterxml.jackson.databind.JsonMappingException: Infinite recursion (StackOverflowError) (through reference chain: org.locationtech.jts.geom.Polygon["envelope"]->org.locationtech.jts.geom.Polygon["envelope"]->...)` 

Envelope objects don't themselves have a field or method for recursively getting their envelope. But Geometries do have a getEnvelope method that for some reason returns a Geometry rather than an Envelope. So the return value of Geometry.getEnvelope() is another Geometry, and it's turtles all the way down. We should just filter Geometry.envelope out of the serialization, since they're lazy-initialized on request anyway.

Adding a `@JsonIgnore` annotation on the mergedBuffers field of the FeedStatistic class eliminates the problem, but I haven't had any luck trying to just filter out the problematic envelope property with a `@JsonIgnoreProperties({"envelope", "precisionModel", "factory"})` on the same field, probably because its sub-geometries also have recursive envelopes.

abyrd:
@landonreed another thing I just noticed: in the serialized feed stats, we have: 

`"datesOfService" : [ "2018-12-30", "2018-12-30", "2018-12-30", "2018-12-31", "2018-12-31", "2018-12-31", "2018-12-31", "2018-12-31"` ...

Note that some dates are repeating over and over.

landonreed:
Closing in favor of #290.

PR #104, opened 2018-03-07
Allow null agency_id; fix calendar_dates false duplicate service_id

landonreed:
This fixes the following:
- [Per the GTFS spec](https://developers.google.com/transit/gtfs/reference/#agencytxt), agency ID can be empty if there is only one agency in the feed.  This fix allows this empty value when editing agencies iff there is one agency.
- Multiple calendar_dates rows with the same `service_id` value should not cause duplicate ID errors to be recorded.
- Allow empty `transfers` field fare_attributes table
- Handle null ints when filling values in `EntityPopluator`

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/Table.java
+++ b/src/main/java/com/conveyal/gtfs/loader/Table.java
@@ -43,6 +43,8 @@ public class Table {
     private Table parentTable;
     /** When snapshotting a table for editor use, this indicates whether a primary key constraint should be added to ID. */
     private boolean usePrimaryKey = false;
+    /** Indicates whether the table has unique key field. */
+    private boolean hasUniqueKeyField = true;
 
     public Table (String name, Class<? extends Entity> entityClass, Requirement required, Field... fields) {
         this.name = name;
@@ -113,7 +115,7 @@ public class Table {
         new DateField("feed_start_date", OPTIONAL),
         new DateField("feed_end_date", OPTIONAL),
         new StringField("feed_version", OPTIONAL)
-    );
+    ).keyFieldIsNotUnique();
 
     public static final Table ROUTES = new Table("routes", Route.class, REQUIRED,
         new StringField("route_id",  REQUIRED),
@@ -137,8 +139,9 @@ public class Table {
             // FIXME: referential integrity check for zone_id for below three fields?
             new StringField("origin_id", OPTIONAL),
             new StringField("destination_id", OPTIONAL),
-            new StringField("contains_id", OPTIONAL)
-    ).withParentTable(FARE_ATTRIBUTES).addPrimaryKey();
+            new StringField("contains_id", OPTIONAL))
+            .withParentTable(FARE_ATTRIBUTES)
+            .addPrimaryKey().keyFieldIsNotUnique();
 
     public static final Table SHAPES = new Table("shapes", ShapePoint.class, OPTIONAL,
             new StringField("shape_id", REQUIRED),
@@ -197,8 +200,9 @@ public class Table {
             new StringField("from_stop_id", REQUIRED).isReferenceTo(STOPS),
             new StringField("to_stop_id", REQUIRED).isReferenceTo(STOPS),
             new StringField("transfer_type", REQUIRED),
-            new StringField("min_transfer_time", OPTIONAL)
-    ).addPrimaryKey();
+            new StringField("min_transfer_time", OPTIONAL))
+            .addPrimaryKey()
+            .keyFieldIsNotUnique();
 
     public static final Table TRIPS = new Table("trips", Trip.class, REQUIRED,
         new StringField("trip_id",  REQUIRED),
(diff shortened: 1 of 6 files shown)


landonreed:
OK @abyrd, this is ready for your review

Tristramg:
:+1: I just ran into this problem. Thank you!

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -746,37 +747,38 @@ public void checkReferencesAndUniqueness(String keyValue, int lineNumber, Field
...
-                }
-            } else {
-                listOfUniqueIds = referenceTracker.transitIds;
-                uniqueId = transitId;
             }
             // Add ID and check duplicate reference in entity-scoped IDs (e.g., stop_id:12345)
-            boolean isDuplicate = !listOfUniqueIds.add(uniqueId);
-            if (isDuplicate) {
+            boolean valueAlreadyExists = !listOfUniqueIds.add(uniqueId);
+            if (!this.name.equals("calendar_dates") && valueAlreadyExists) {
```
abyrd: It seems strange to have a hard coded single exception here. This code is already inside a conditional `field.name.equals(uniqueKeyField)`. Do we actually need to flag this field as the "unique key field" in `calendar_dates` when it's not unique? Is it even a good idea to do so?

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -746,37 +747,38 @@ public void checkReferencesAndUniqueness(String keyValue, int lineNumber, Field
...
-            // whether an order field is part of the "unique ID."
-            String transitId = String.join(":", getKeyFieldName(), keyValue);
-            Set<String> listOfUniqueIds;
-            String uniqueId;
             // Check for duplicate IDs and store entity-scoped IDs for referential integrity check
             if (isOrderField) {
                 // Check duplicate reference in set of field-scoped id:sequence (e.g., stop_sequence:12345:2)
                 // This should not be scoped by key field because there may be conflicts (e.g., with trip_id="12345:2"
                 listOfUniqueIds = referenceTracker.transitIdsWithSequence;
                 uniqueId = String.join(":", field.name, keyValue, string);
```
abyrd: This shortcut using string concatenation is only correct most of the time, and if the constituent strings never contain colons.  I think this might be what you are referring to by "should not be scoped by key field because there may be conflicts" because you cite a trip_id containing a colon character. We should probably use a Tuple type (which don't exist in Java but get defined in libraries and places in our own code). Or just a List of String. Though these things are going to get huge for stop_times tables.

landonreed: Can you name a case where this would not work?  Are you suggesting that the `listofUniqueIds` should be a List rather than a Set (also perhaps the variable is not well-named).

src/main/java/com/conveyal/gtfs/loader/Table.java
landonreed: I think I see what you mean.  `service_id` should not be the "unique key field" for `calendar_dates` in the first place.  So when we set the value of `uniqueKeyField`, we could do something like the following: 
```
String uniqueKeyField = this.name.equals("calendar_dates") 
? null
: orderField != null ? orderField : keyField;
```

This still hard codes the calendar_dates exception, but the meaning is a little clearer.

abyrd: That's basically what I mean, but I was thinking one level even further up. `Table.getKeyFieldName` always returns the first field in the table, but several tables don't have a unique key (nor a compound key using a sequence number). There are several tables that will create this same situation. Scanning through the table definitions, I see transfers, frequencies, feed_info, and fare_rules where the first field does not serve as a unique key, neither alone nor in combination with a sequence number. For those tables, I'd say getKeyField should return null. We might need another fluent method to set the key field (or de-set it) on these tables.

landonreed: Ah, right.  Yes, that makes sense to me.

src/main/java/com/conveyal/gtfs/loader/Table.java
abyrd: Hmm, I guess it may always work since tables with an order field produce strings that are guaranteed to end in a colon followed by an integer. But I don't fully understand why these are being scoped by the name of the key field. Aren't all the identifiers coming from the same table part of the same namespace? Why would we be checking strings from different tables within a single reference tracker? I will need to get into this more deeply to understand how it works.

abyrd: Oh, I see you're checking cross-table references and uniqueness of keys all at the same time. I get the sense this could be refactored to be much more readable. There are also terms used in the comments like `transit ID` which I don't believe are defined anywhere. Maybe we can go over this together some time and try to clean it up. It does seem like it would properly perform the needed checks, but we need to improve the maintainability/readability of this section.

abyrd:
@landonreed, I got a question from @Tristramg (now working with Etalab) about this commit. Is it expected to be a breaking change with respect to data already loaded into the RDBMS?

landonreed:
@abyrd and @Tristramg, I don't think this is a breaking change.  I believe all this branch does is allow a `null` agency ID when an agency entity is being edited and handle properly the validation of unique keys (i.e., squash previous validation errors for fields that are not unique like `calendar_dates#service_id`.).

Tristramg:
Hi,

Indeed, I tested and it works. Not sure what happened when I tried it
previously. I probably wasn’t awake.

Anyways, thanks a lot for the fix

PR #108, opened 2018-03-29
Add bounding box and string search GraphQL queries; Reconcile stop times to pattern stop changes

landonreed:
### Bounds: `maxLat`, `maxLon`, `minLat`, `minLon`
This PR adds support for GraphQL-based bounding box queries for both GTFS stops and patterns.  It does this by simply chaining conditions onto the where clause constructed for the SQL query.

### String search: `search: "Some String"`
It also adds string search for stops (on stop_id, stop_code, and stop_name) and routes (route_id, route_short_name, and route_long_name).  A column metadata query ensures that the route or stop table has the respective search fields before issuing a query to them.

## Bug fixes
This adds missing code to reconcile stop times to any changes to pattern stops (i.e., additions, deletions, transpositions) as well as linking pattern stops fields to stop times (e.g., timepoint, pickup or drop off).  Other tables with redundant data are also "linked", including trips which inherits `wheelchair_accessible` from routes and `direction_id` from patterns.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
+++ b/src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
@@ -255,12 +255,12 @@ public class GraphQLGtfsSchema {
             .field(MapFetcher.field("route_desc"))
             .field(MapFetcher.field("route_url"))
             .field(MapFetcher.field("route_branding_url"))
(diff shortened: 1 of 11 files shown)


landonreed:
@abyrd, this is ready for your review, but unfortunately I need to merge now in order to deploy with an updated data tools jar.

PR #110, opened 2018-04-12
Ignore 0 travel time for very close stops

Tristramg:
We have lots of validations errors because timetables are rounded to the minute.

This would allow to reduce the noise in the validation results


landonreed:
@Tristramg, I agree this can become very noisy.  I like the solution you have proposed.  @abyrd?

abyrd:
@Tristramg I agree this can be a problem. These validations produce a ton of noise on many feeds, for exactly the reason you mentioned: it's common to round off timetables. I consider that to be a warning-worthy characteristic of GTFS feeds, but one that is nonetheless common and tolerable.

I would suggest instead that we attempt to detect rounded off feeds (i.e. all seconds are 00) and on those feeds we register a single feed-wide warning about rounding, and only on feeds with that characteristic we would disable zero-time hops.

Alternatively on a per-hop basis, we can disable the warning if the starting and ending stop_times end in 00 seconds and imply a realistic upper bound on speed (as you checked with the 300m limit).

landonreed:
Closing in favor of #145.

PR #111, opened 2018-04-16
Filter GTFS entities on date/period of time for GraphQL fetch

landonreed:
This PR adds additional GraphQL arguments for `trips` to filter them by date and time period (i.e., trips that start between 9am and 12pm).  

The `JDBCFetcher` uses the arguments to build SQL fragments that draw from the `service_dates` table.  So these arguments are currently only functional for feeds that have gone through validation and have not been altered since validation (i.e., these arguments will not produce accurate results for editor feeds).


src/main/java/com/conveyal/gtfs/graphql/fetchers/JDBCFetcher.java
```diff
@@ -42,12 +45,19 @@
     public static final String LIMIT_ARG = "limit";
     public static final String OFFSET_ARG = "offset";
     public static final String SEARCH_ARG = "search";
+    public static final String DATE_ARG = "date";
```
abyrd: Is it really necessary to define symbolic constants that have exactly the same name as their own contents? I guess the advantage is that if they're reused in several places, there's no risk of a misspelling in one place, since they're checked by the IDE and compiler.

src/main/java/com/conveyal/gtfs/graphql/fetchers/JDBCFetcher.java
```diff
@@ -189,9 +199,6 @@ public static GraphQLFieldDefinition field (String tableName) {
         for (String key : argumentKeys) {
             // Limit and Offset arguments are for pagination. projectStops is used to mutate shape points before
             // returning them. All others become "where X in A, B, C" clauses.
-            Set<String> argsToSkip = new HashSet<>();
```
abyrd: The comment above only mentions pagination and projectStops, saying that all others become "where" clauses. In fact it looks like we're now skipping paginationArgs, boundingBoxArgs, and dateTimeArgs. The comment should be updated.

src/main/java/com/conveyal/gtfs/graphql/fetchers/JDBCFetcher.java
landonreed: Yes, reusability was my reasoning here. If the arguments need to be added to other types, it's also much more convenient and less error prone to type out a constant than a string literal in the IDE.

src/main/java/com/conveyal/gtfs/graphql/fetchers/JDBCFetcher.java
landonreed: This comment is actually referencing the old file in the diff.  However, the argsToSkip variable in the new file is lacking a good comment, so I will update it.

src/main/java/com/conveyal/gtfs/graphql/fetchers/JDBCFetcher.java
```diff
@@ -230,6 +242,25 @@ public static GraphQLFieldDefinition field (String tableName) {
...
+        if (argumentKeys.containsAll(dateTimeArgs)) {
+            // Handle filtering stop_times by date and period of time (e.g., 4/12/2018 8:00am-11:00am). NOTE: time args
+            // are actually specified in seconds since midnight.
+            // NOTE: This query must be run only on feeds that have run through validation, which generates the
+            // service_dates table. In other words, feeds generated by the editor cannot be queried with the date/time args.
+            fromTables.add(String.format("%s.trips", namespace));
+            String date = (String) arguments.get(DATE_ARG);
+            // Gather all service IDs that run on the provided date.
+            fromTables.add(String.format("(select distinct service_id from %s.service_dates where service_date = '%s') as unique_service_ids_in_operation",
+                    namespace, date));
```
evansiroky: This is prone to SQL injection.

landonreed:
OK, SQL the injection vulnerabilities in `JdbcFetcher` have been repaired.

PR #115, opened 2018-05-04
auto increment error ID in favor of using count

landonreed:
This PR:
- replaces usage of `errorCount` field to track IDs with a serial, auto-incrementing ID.
- removes the `errorCount` field `SQLErrorStorage` in favor of making a SQL query for `count(*)` in `getErrorCount()`

This change was added to prevent the possibility for the ID becoming out of sync and causing issues like catalogueglobal/datatools-ui#111.


landonreed:
Closed in favor of #116.  However, after discussion with @abyrd, we have decided to keep the SQL query in `getErrorCount`, but to revert to the use of `max(id)` to determine the next error ID.

PR #120, opened 2018-06-05
More tests

evansiroky:
* Adds a lot more tests to verify core functionality of gtfs-lib with respect to SQL storage


landonreed:
@evansiroky, could you isolate your commits from the others before we merge this?

evansiroky:
This build is intended to fail because of a failing test that checks for SQL injection.  Those issues will be fixed in the sql-sanitizing branch.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -0,0 +1,142 @@
...
+        JdbcTableWriter deleteTableWriter = new JdbcTableWriter(Table.FEED_INFO, testDataSource, testNamespace);
+        int deleteOutput = deleteTableWriter.delete(
+            createdFeedInfo.id,
+            true
+        );
+        LOG.info("delete output:");
+        LOG.info(updateOutput);
+
+        // make sure record does not exist in DB
+        String sql = "select * from " + testNamespace + "." + Table.FEED_INFO.name +
```
landonreed: I've begun trying to use `String.format` more consistently in this codebase.  Could you replace instances where strings are built with `+`?

src/test/resources/fake-agency-interpolated-stop-times/agency.txt
```diff
@@ -0,0 +1,2 @@
+agency_id,agency_name,agency_url,agency_lang,agency_phone,agency_email,agency_timezone,agency_fare_url,agency_branding_url
```
landonreed: Is there a reason that you have unzipped folders instead of zip files?  Is this so that the files are more easily version controlled?  Also, why is the `fake-agency-interpoloated-stop-times.zip` tracked but `fake-agency.zip` is not?

src/test/java/com/conveyal/gtfs/GTFSFeedTest.java
```diff
@@ -152,4 +169,114 @@ public void canDoRoundtripLoadAndWriteToZipFile() throws IOException {
...
+     * @throws GTFSFeed.FirstAndLastStopsDoNotHaveTimes
+     */
+    @Test
+    public void canGetInterpolatedTimes() throws GTFSFeed.FirstAndLastStopsDoNotHaveTimes, IOException {
+        String tripId = "a30277f8-e50a-4a85-9141-b1e0da9d429d";
+
+        String gtfsZipFileName = TestUtils.zipFolderFiles("fake-agency-interpolated-stop-times");
+
+        GTFSFeed feed = GTFSFeed.fromFile(gtfsZipFileName);
+        Iterable<StopTime> stopTimes = feed.getInterpolatedStopTimesForTrip(tripId);
```
landonreed: I'm not even sure if this method is really used anywhere anymore.  It might not be worth developing a test for.  @abyrd, is this used anywhere in r5?

src/test/resources/fake-agency-interpolated-stop-times/agency.txt
evansiroky: Yes, it's so they can be more easily tracked in version control.  I'll get rid of that other zip file, it shouldn't be there.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
evansiroky: Will do

src/test/java/com/conveyal/gtfs/GTFSFeedTest.java
evansiroky: If something isn't used anymore we should remove it from the codebase.

PR #122, opened 2018-06-05
Update use_frequency field for patterns on editor snapshot

### LINKED ISSUE #121: Patterns imported for editing do not have use_frequency field set properly
When importing a GTFS file with frequencies entries, the editor-specific pattern field `use_frequency` is not updated according to the trips that reference it, making the frequencies uneditable in the UI.

landonreed:
Fixes #121.

Note: this updates patterns in the editor that have at least one referencing frequency trip to be "frequency-based."  What should be done if there are both frequency- and timetable-based trips referencing a single pattern?

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsSnapshotter.java
+++ b/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsSnapshotter.java
@@ -193,6 +193,22 @@ public class JdbcGtfsSnapshotter {
             int calendarsUpdated = statement.executeUpdate(updateOtherSql);
             LOG.info("Updated description for {} calendars", calendarsUpdated);
         }
+        if (Table.TRIPS.name.equals(table.name)) {
+            // Update use_frequency field for patterns. This sets all patterns that have a frequency trip to use
+            // frequencies. NOTE: This is performed after copying the TRIPS table rather than after PATTERNS because
+            // both tables (plus, frequencies) need to exist for the successful operation.
+            // TODO: How should this handle patterns that have both timetable- and frequency-based trips?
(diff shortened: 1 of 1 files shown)


src/main/java/com/conveyal/gtfs/loader/JdbcGtfsSnapshotter.java
```diff
@@ -193,6 +193,22 @@ private void populateDefaultEditorValues(Connection connection, String tablePref
...
+            // Update use_frequency field for patterns. This sets all patterns that have a frequency trip to use
+            // frequencies. NOTE: This is performed after copying the TRIPS table rather than after PATTERNS because
+            // both tables (plus, frequencies) need to exist for the successful operation.
+            // TODO: How should this handle patterns that have both timetable- and frequency-based trips?
+            // NOTE: The below substitution uses relative indexing. All values "%<s" reference the same arg as the
+            // previous format specifier (i.e., tablePrefix).
+            String updatePatternsSql = String.format(
+                    "update %spatterns set use_frequency = 1 " +
+                    "from (select distinct %<strips.pattern_id from %<strips, %<sfrequencies where %<sfrequencies.trip_id = %<strips.trip_id) freq " +
+                    "where freq.pattern_id = %<spatterns.pattern_id",
```
evansiroky: I think an [argument index](http://java2s.com/Tutorials/Java/String/How_to_use_an_argument_index_to_specify_which_to_use_in_Java_Formatter.htm#Using_an_Argument_Index_) would be a more readable way for this string substitution.

landonreed: Yea, perhaps you're right.

PR #126, opened 2018-06-13
GraphQL testing and Sql sanitizing

evansiroky:
- Write tests for GraphQL queries
- Write tests for JDBCTableWriter
- Fix a SQL injection vulnerability in JDBCTableWriter


src/test/java/com/conveyal/gtfs/graphql/GTFSGraphQLTest.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.Map;
+
+import static com.conveyal.gtfs.GTFS.createDataSource;
+import static com.conveyal.gtfs.GTFS.load;
+import static com.conveyal.gtfs.GTFS.validate;
+import static com.conveyal.gtfs.TestUtils.getResourceFileName;
+import static com.zenika.snapshotmatcher.SnapshotMatcher.matchesSnapshot;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class GTFSGraphQLTest {
```
abyrd: Please add Javadoc to all new classes explaining at least briefly what they do (in this case what sort of tests they contain).

src/test/java/com/conveyal/gtfs/graphql/GTFSGraphQLTest.java
```diff
@@ -0,0 +1,229 @@
...
+     */
+    private Map<String, Object> queryGraphQL(String queryFilename) throws IOException {
+        Map<String, Object> variables = new HashMap<String, Object>();
+        variables.put("namespace", testNamespace);
+        return queryGraphQL(queryFilename, variables, testDataSource);
+    }
+
+    /**
+     * Helper method to execute a GraphQL query and return the result
+     */
```
abyrd: Please clarify in Javadoc what the parameters do, most importantly queryFilename. It's not obvious in the code that calls this method that the filenames are test fixtures containing the full query text.

src/test/java/com/conveyal/gtfs/util/UtilTest.java
```diff
@@ -26,4 +30,34 @@ public void canHumanize() {
...
+
+    private void testNamespace(String namespace, boolean shouldFail) {
+        boolean errorThrown = false;
+        try {
+            ensureValidNamespace(namespace);
+        } catch (IllegalStateException e) {
+            errorThrown = true;
+            if (shouldFail) {
+                assertThat(e.getMessage(), equalTo(
+                    "Namespace must only have alphanumeric characters or the underscore symbol"
```
abyrd: This is testing for exact implementation details (the exact contents of an error message) instead of general behavior (it fails when input is not clean). We should just check for an exception, and possibly check the type of exception (in which case maybe we should define our own exception type, e.g. PossibleSQLInjectionException).

Another option is to perform all SQL injection testing with methods that return unambiguous booleans (acceptable / non-acceptable) instead of exceptions. Then we can just test that boolean value, which purely concentrates on the behavior and meaning instead of implementation.

evansiroky:
I see your concern about the stops that don't necessarily require an order to them.  So far these tests have been consistent, but if they start becoming flaky because of that ordering issue it may be a better idea to verify the output against a JSON Schema or something.

abyrd:
I think it's a good idea to verify the contents rather than just the schema. But the contents need to be verified in an order-independent way. I would expect a library that checks JSON snapshots to be able to handle unordered elements. If it doesn't, this is a shortcoming and should at least be mentioned in the Javadoc so someone doesn't spend a lot of time trying to figure out why tests fail in the future.

evansiroky:
Ok, I added a comment about the ordering in the GraphQLTest file.

landonreed:
@evansiroky's latest commits appear to address the requested changes. I'm going to merge this.

PR #133, opened 2018-08-23
Editor: Allow null values for foreign ref fields or required fields where permitted

landonreed:
This fixes a couple of editor bugs:
- where create/update requests with foreign refs (such as `fare_rules#route_id`)
could not be null even if the table definition permits the null value (catalogueglobal/datatools-ui#224) and
- where required fields that permit empty values (`fare_attributes#transfers`) were not allowed in create/update requests (catalogueglobal/datatools-ui#237)

It also adds a basic CRUD test for fares.


evansiroky:
I'll assign this back to you, Landon, and let you decide if this is a good time to start adding tests.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -140,6 +148,166 @@ public void canPreventSQLInjection() throws IOException, SQLException, InvalidNa
...
+        assertThat(updatedFareDTO.fare_rules[0].fare_id, equalTo(updatedFareId));
+
+        // try to delete record
+        JdbcTableWriter deleteTableWriter = createTestTableWriter(fareTable);
+        int deleteOutput = deleteTableWriter.delete(
+                createdFare.id,
+                true
+        );
+        LOG.info("delete {} output:", fareTable.name);
+        LOG.info(updateOutput);
```
evansiroky: This should probably be `deleteOutput`.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -140,6 +148,166 @@ public void canPreventSQLInjection() throws IOException, SQLException, InvalidNa
...
+        // make sure fare_rules record does not exist in DB
+        String fareRulesSql = String.format(
+                "select * from %s.%s where id=%d",
+                testNamespace,
+                Table.FARE_RULES.name,
+                createdFare.fare_rules[0].id
+        );
+        LOG.info(fareRulesSql);
+        ResultSet fareRulesResultSet = testDataSource.getConnection().prepareStatement(fareRulesSql).executeQuery();
+        assertThat(fareRulesResultSet.getFetchSize(), equalTo(0));
```
evansiroky: Now that I look at these tests, it seems like they could really use a sql check to make sure that the data was created and updated.  Also, these 2 lines could be made into a reusable method `assertThatSqlQueryYieldsZeroRows` or something like that.

evansiroky:
Excellent!  I gave an idea of how to improve the CRUD tests, but if that seems out of scope for this PR, let's create an issue for that and do that later.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
landonreed: Good idea.

landonreed: Hmm, actually I think the checking for created/updated data gets a little muddied because we would then need to ensure that we're using the same connection to see the data we expect. I think the first part of your comment is out of scope.

PR #134, opened 2018-08-28
Delete feed from RDBMS

landonreed:
This PR adds the ability to delete a feed (essentially dropping a schema) to the RDBMS-backed code path.

It also updates the documentation for the cmd line interface (in addition to adding the `--delete` and `--json` options).

Finally, it tweaks the `TableLoadResult` and `FeedLoadResult` objects to include fields for `fileSize` and `filename`, respectively.


src/main/java/com/conveyal/gtfs/GTFS.java
```diff
@@ -159,12 +202,17 @@ public static void main (String[] args) {
...
-        if (!(cmd.hasOption("export") || cmd.hasOption("snapshot") || cmd.hasOption("load") || cmd.hasOption("validate") || cmd.hasOption("graphql"))) {
-            LOG.error("Must specify one of 'snapshot', 'load', 'validate', 'export', or 'graphql'.");
+        if (!(cmd.hasOption("export") || cmd.hasOption("snapshot") || cmd.hasOption("load") || cmd.hasOption("validate") || cmd.hasOption("delete"))) {
+            LOG.error("Must specify one of 'snapshot', 'load', 'validate', 'export', or 'delete'.");
             printHelp(options);
             return;
         }
-
+        boolean storeResults = cmd.hasOption("json");
+        File directory = null;
```
abyrd: I would move the creation of the objectMapper down here, conditional upon storeResults=true.

src/main/java/com/conveyal/gtfs/GTFS.java
```diff
@@ -229,12 +292,21 @@ public static void main (String[] args) {
...
-        if (cmd.hasOption("graphql")) {
-            Integer port = Integer.parseInt(cmd.getOptionValue("graphql"));
-            LOG.info("Starting GraphQL server on port {}", port);
-            throw new UnsupportedOperationException();
-        }
+        if (cmd.hasOption("delete")) {
+            String namespaceToDelete = cmd.getOptionValue("delete");
 
+            if (namespaceToDelete != null) {
+                LOG.info("Exporting feed with unique identifier {}", namespaceToDelete);
```
abyrd: Message appears to be incorrect.

src/main/java/com/conveyal/gtfs/GTFS.java
```diff
@@ -252,7 +324,7 @@ private static Options getOptions () {
                 .desc("zip file path for the exported GTFS").build());
         options.addOption(Option.builder().longOpt("load").hasArg()
                 .argName("file").desc("load GTFS data from the given file").build());
-        options.addOption(Option.builder().longOpt("validate").hasArg().optionalArg(true).argName("feed")
+        options.addOption(Option.builder().longOpt("validate").hasArg().optionalArg(true).argName("feedId")
```
abyrd: "feedId" has another meaning distinct from our unique-IDs for loaded feed versions ("namespaces"). We should probably avoid use of the word feedId.

landonreed: I would agree with that.  However, it has been used in a number of places within this class already to mean `namespace` often with a parenthetical clarifying that it is referring to the schema identifier.  I will simply update all of the cmd line option args/text for now.

src/main/java/com/conveyal/gtfs/GTFS.java
```diff
@@ -179,8 +179,8 @@ public static DataSource createDataSource (String url, String username, String p
      * It also lets you run a GraphQL API for all the feeds loaded into the database.
      */
     public static void main (String[] args) throws IOException {
-        // Object mapper used for writing load or validation results to file.
-        ObjectMapper mapper = new ObjectMapper();
+        // Object mapper used for writing load or validation results to file (instantia).
```
abyrd: There appears to be stray text in this comment. This is a detail, but I'd find the code easier to read if the variable were declared and initialized to null just before it is conditionally set to a new instance, around line 211 where we also declare `File directory = null;`

landonreed:
@abyrd, I hadn't changed anything related to the use of `feedId` because I was awaiting a response to the above comment: https://github.com/conveyal/gtfs-lib/pull/134#discussion_r216766478.

src/main/java/com/conveyal/gtfs/GTFS.java
abyrd: I don't necessarily think that we should change this right away, now that I see how many times the word feedId is used in this class. But I don't see why it would be a breaking change. The validate tasks's argument is called `feed` rather than `feedId` so the command line options are already a bit incoherent. But I don't even think the argument names are used anywhere. We'd want to also change some parameter names, but since parameters are positional in Java it wouldn't break callers.

I will make a ticket for this if we're not going to change it soon.

abyrd:
Oh, didn't see that comment before. I just replied. We can always do it later, maybe just make an issue to clean this up later.

src/main/java/com/conveyal/gtfs/GTFS.java
landonreed: Oh, you're right.  Changing the name `feedId` wouldn't affect any command line instructions, just the output of the help text.  I will update these arg names here and we can use the ticket you create to track down/update other inconsistent usages.

landonreed:
I'm going to go ahead and merge this.  I've created a ticket to track the namespace/feedId conflation issue here: #142.

PR #136, opened 2018-08-29
v4.0.0

landonreed:
This release includes create, update, and delete editing capabilities for feeds stored in RDBMS.

PR #137, opened 2018-08-29
v4.0.0

landonreed:
This fixes the travis configuration and will be the *real* release of v4.0.0 (see #136 for previous attempt).


landonreed:
I can confirm that the Travis configuration looks how we want it to:
image

PR #141, opened 2018-09-12
Export improvements, proper schedule_exception handling and more tests

### LINKED ISSUE #140: Round double fields in GTFS Export to save space
In doing a roundtrip of loading a GTFS and then exporting it using gtfs-lib, it appears that double fields (ie fare_attributes.price and shape lat/lons) have unneeded precision exported to GTFS files.

Here's a before and after of fare_attributes.txt:

Before:

```
fare_id,price,currency_type,payment_method,transfers,transfer_duration
local_ashland,1.23,USD,0,1,5400

```

After:

```
fare_id,price,currency_type,payment_method,transfers,transfer_duration
std_fare,1.22999999999999998,USD,0,1,5400
```

And here is a before and after of shapes.txt:

Before: 

```
shape_id,shape_pt_lat,shape_pt_lon,shape_pt_sequence
21575188-6616-47aa-b224-a5132d39195f,42.357651,-122.889112,17
35bfb55e-51f5-4eaa-b928-c653e53671af,42.341635664799,-122.876885257774,48
ab16ab9d-b0ae-4e70-8ec0-ee65421463fe,42.322334,-122.873422,2
93ce80ab-f9b9-4872-9365-59838827277c,42.31277,-122.87831,11
a2f6e565-27d5-4701-9a5d-ec92f7aa58f2,42.438609596408,-122.836868037991,7
536da693-4755-4fb8-b819-13d199494b55,42.2808500993758,-122.823298741776,32
868c0b0b-4139-4fb5-b018-ba0d2d7d61e9,42.3427474405019,-122.878438256886,44
536da693-4755-4fb8-b819-13d199494b55,42.185328584035,-122.691527169438,82
a2f6e565-27d5-4701-9a5d-ec92f7aa58f2,42.388402,-122.856084,35
a2f6e565-27d5-4701-9a5d-ec92f7aa58f2,42.342132,-122.860172,57
```

After:

```
shape_id,shape_pt_sequence,shape_pt_lat,shape_pt_lon,shape_dist_traveled
21575188-6616-47aa-b224-a5132d39195f,17,42.3576509999999971,-122.889111999999997,
35bfb55e-51f5-4eaa-b928-c653e53671af,48,42.341635664799,-122.876885257774006,
ab16ab9d-b0ae-4e70-8ec0-ee65421463fe,2,42.3223339999999979,-122.873422000000005,
93ce80ab-f9b9-4872-9365-59838827277c,11,42.3127700000000004,-122.878309999999999,
a2f6e565-27d5-4701-9a5d-ec92f7aa58f2,7,42.4386095964079999,-122.836868037990996,
536da693-4755-4fb8-b819-13d199494b55,32,42.2808500993757974,-122.823298741776,
868c0b0b-4139-4fb5-b018-ba0d2d7d61e9,44,42.3427474405019026,-122.878438256886,
536da693-4755-4fb8-b819-13d199494b55,82,42.1853285840349983,-122.691527169438004,
a2f6e565-27d5-4701-9a5d-ec92f7aa58f2,35,42.3884019999999992,-122.856083999999996,
a2f6e565-27d5-4701-9a5d-ec92f7aa58f2,57,42.3421319999999994,-122.860172000000006,
```

Seems like we should always be rounding the currency, or perhaps using a currency formatter.  And for the shapes maybe those should be rounded at 6 decimal places since [any precision beyond that is unneeded](https://gis.stackexchange.com/a/109178/12590).

evansiroky:
### Features & Fixes
- Round outputted double fields to avoid excessive precision.  (Fixes #140)
- Properly store calendar_dates as schedule_exceptions (Refs https://github.com/catalogueglobal/datatools-ui/issues/138)
- Make sure TimeFields are always exported in HH:MM:SS format
- Make sure import/export roundtrip of a GTFS that omits the calendar.txt file can work properly

### Additional Tests
- Add tests for exporting from the database to a GTFS zip file.
- Add test for importing and exporting a GTFS zip file that omits calendar.txt


src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -478,13 +503,48 @@ public String generateSelectSql (String namespace, Requirement minimumRequiremen
...
     }
 
+    /**
+     * Generate a select statement from the columns that actually exist in the database table
+     */
+    public String generateSelectAllExistingFieldsSql(Connection connection, String namespace) throws SQLException {
+        // select all columns from table
+        // FIXME This is postgres-specific and needs to be made generic for non-postgres databases.
+        PreparedStatement statement = connection.prepareStatement(
+            "SELECT column_name FROM information_schema.columns WHERE table_schema = ? AND table_name = ?"
```
landonreed: I have a concern about this method.  I'm not sure it's actually exporting whatever was loaded into the database from the GTFS feed.  It looks it's filtering the columns exported by fields it finds in the table definition. This means it would drop any extensions that are not a part of that definition.  This needs to be addressed.

We probably also need a test case to ensure that non-editor feeds do not lose extensions after a round trip (import -> export).

src/test/java/com/conveyal/gtfs/TestUtils.java
```diff
@@ -35,7 +35,11 @@ public static void dropDB(String dbName) {
             "AND pid <> pg_backend_pid()", dbName
         ));
         // drop the db
-        executeAndClose(String.format("DROP DATABASE %s", dbName));
+        if(executeAndClose(String.format("DROP DATABASE %s", dbName))) {
```
landonreed: Something I just thought of: what if someone has a database already for the DB name that gets auto-generated?  The tests won't try to use that and/or drop it upon completion, will they?  Just making sure...

landonreed:
One other thing that I just remembered.  datatools-server used to enforce [only one schedule exception per date](https://github.com/catalogueglobal/datatools-server/blob/dev_2.0.0/src/main/java/com/conveyal/datatools/editor/controllers/api/ScheduleExceptionController.java#L104).  The UI still does enforce that constraint.  So I think we may want to reconsider whether the calendar_dates -> schedule_exceptions mapping is one-to-one or many-to-one (e.g., for July 4th add service A and remove service B becoming one schedule_exception instead of two) to enforce this rule. cc @abyrd

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -155,9 +156,9 @@ public Table (String name, Class<? extends Entity> entityClass, Requirement requ
     public static final Table SHAPES = new Table("shapes", ShapePoint.class, OPTIONAL,
             new StringField("shape_id", REQUIRED),
             new IntegerField("shape_pt_sequence", REQUIRED),
-            new DoubleField("shape_pt_lat", REQUIRED, -80, 80),
-            new DoubleField("shape_pt_lon", REQUIRED, -180, 180),
-            new DoubleField("shape_dist_traveled", OPTIONAL, 0, Double.POSITIVE_INFINITY),
+            new DoubleField("shape_pt_lat", REQUIRED, -80, 80, 6),
+            new DoubleField("shape_pt_lon", REQUIRED, -180, 180, 6),
+            new DoubleField("shape_dist_traveled", OPTIONAL, 0, Double.POSITIVE_INFINITY, 2),
```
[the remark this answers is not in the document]
evansiroky: Oh interesting, I just assumed the units were meters, but I didn't realize they could be literally any kind of distance measure.  In this case, we shouldn't have any limitation on precision for this field because it is theoretically possible to use Astronomical Units.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -299,12 +343,22 @@ private void cleanUpZipFile() {
...
-    private TableLoadResult export (Table table) {
+    private TableLoadResult export (Table table, Connection connection) {
         if (fromEditor) {
             // Default behavior for exporting editor snapshot tables is to select only the spec fields.
             return export(table, table.generateSelectSql(feedIdToExport, Requirement.OPTIONAL));
         } else {
-            return export(table, null);
+            String existingFieldsSelect = null;
+            try {
+                existingFieldsSelect = table.generateSelectAllExistingFieldsSql(connection, feedIdToExport);
```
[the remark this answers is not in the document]
evansiroky: I added more comments in various places.  Yeah, filterSql probably shouldn't be null or at least an error should be thrown if it is.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsSnapshotter.java
```diff
@@ -141,17 +149,169 @@ private TableLoadResult copy (Table table, boolean createIndexes) {
...
+        }
+
+        if (scheduleExceptionsTableExists) {
+            // schedule_exceptions table already exists in namespace being copied from.  Therefore, we simply copy it.
+            return copy(Table.SCHEDULE_EXCEPTIONS, true);
+        } else {
+            // schedule_exceptions does not exist.  Therefore, we generate schedule_exceptions from the calendar_dates.
+            TableLoadResult tableLoadResult = new TableLoadResult();
+            try {
+                String sql = String.format(
```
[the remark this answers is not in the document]
evansiroky: Ah, `createSqlTable` will work, but the `generateInsertSql` isn't specific enough, so I think I still need to create my own statement.

src/main/java/com/conveyal/gtfs/loader/Table.java
evansiroky: I'm not sure I follow.  This was added to mimic existing behavior where all of the database fields were exported in the fashion of `select * from table`.  So the past way of doing things did basically the same thing except it didn't format certain fields.

src/test/java/com/conveyal/gtfs/TestUtils.java
evansiroky: hmm, well the tests would run into an error because they would try creating a database that already existed.  But yeah, it's a good idea to generate more random database name.

src/main/java/com/conveyal/gtfs/loader/Table.java
landonreed: In this block of code, you're only including the column name if it is found in the canonical list of fields for the table.
https://github.com/conveyal/gtfs-lib/blob/05fe14e560ff81acdf567e02a0b988a7bff4978a/src/main/java/com/conveyal/gtfs/loader/Table.java#L538-L543

If I had a field called `date_color` in the `calendar_dates` table from a feed I imported, on export this field would be dropped.  This is currently only relevant for feeds that are never imported to the editor, where we don't currently guarantee that extensions will make the round trip.

abyrd:
> One other thing that I just remembered: datatools-server used to enforce [only one schedule exception per date](https://github.com/catalogueglobal/datatools-server/blob/dev_2.0.0/src/main/java/com/conveyal/datatools/editor/controllers/api/ScheduleExceptionController.java#L104). The UI still does enforce that constraint. So I think we may want to reconsider whether the calendar_dates -> schedule_exceptions mapping is one-to-one or many-to-one (e.g., for July 4th add service A and remove service B becoming one schedule_exception instead of two) to enforce this rule. cc @abyrd

Just to confirm what @landonreed said on Slack, I agree that if it's always worked this way and the UI enforces it, this is as good a normalization for the incoming exceptions as any - one set of all service IDs added or removed for each date that has any additions/removals.

src/main/java/com/conveyal/gtfs/loader/Table.java
evansiroky: @abyrd I'm don't think this would work as you are suggesting.  Previously and currently, the Java models haven't been involved in the output to csv files.  The writing of the output happens here: https://github.com/conveyal/gtfs-lib/blob/05fe14e560ff81acdf567e02a0b988a7bff4978a/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java#L379-L397
As can be seen there is no middleman in the process, the csv is exported directly from postgres to an outputstream.

abyrd: Thanks @evansiroky I was not familiar with how this works so am just now understanding all the implications. Indeed if we want to preserve all the columns there's no real way to go through the Java model objects.

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -430,20 +454,47 @@ public String generateInsertSql (String namespace, boolean setDefaultId) {
...
                 // wheelchair_accessible in routes and trips). So this filter has been removed.
-                .map(f -> prefix != null // && (f.isForeignReference() || getKeyFieldName().equals(f.name))
+                .map(f -> {
+                    String column = prefix != null // && (f.isForeignReference() || getKeyFieldName().equals(f.name))
                         ? prefix + f.name
-                        : f.name)
+                        : f.name;
+                    // perform translations of certain fields if exporting to a GTFS
+                    if (csvOutput) {
+                        if (DoubleField.class.isInstance(f) && ((DoubleField) f).getOutputPrecision() > 0) {
```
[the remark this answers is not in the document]
evansiroky: oh wow, overriding methods is so much cleaner.

landonreed:
Hmm, @evansiroky, shouldn't the base branch be `dev` not `master`?

evansiroky:
Just changed the target branch to dev and resolved a merge conflict.  This should be ready to be merged.

PR #143, opened 2018-09-26
Fix bad date NPE

landonreed:
Ensure that only non-null dates are processed by the `ServiceValidator` so that we don't cause exceptions when finding first/last dates.


src/main/java/com/conveyal/gtfs/validator/ServiceValidator.java
```diff
@@ -181,6 +181,8 @@ public void complete(ValidationResult validationResult) {
             LocalDate firstDate = LocalDate.MAX;
             LocalDate lastDate = LocalDate.MIN;
             for (LocalDate date : dateInfoForDate.keySet()) {
+                // If the date is invalid, skip.
+                if (date == null) continue;
```
evansiroky: How is this error happening?  Shouldn't the computeIfAbsent method for adding keys and values never yield a null key?  It'd be very helpful to add a failing test case to illustrate what is causing the problem here.

evansiroky:
Reassigning to Landon.  Would like an answer to https://github.com/conveyal/gtfs-lib/pull/143#discussion_r225033222 so I can do a better review.

src/main/java/com/conveyal/gtfs/validator/ServiceValidator.java
landonreed: @evansiroky, there are now tests that load feeds with bad date values. When you run `canLoadFeedWithBadDates`, you'll see that it does encounter a null key.

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -462,8 +536,31 @@ private void assertThatImportedGtfsMeetsExpectations(
                     break;
                 }
             }
-            assertThatPersistenceExpectationRecordWasFound(numRecordsSearched, foundRecord);
+            assertThatPersistenceExpectationRecordWasFound(numRecordsSearched, foundRecord, fieldsWithMismatches);
+        }
+    }
+
+    private void assertThatValidationErrorsMatchExpected(
```
evansiroky: This method is a little unclear as to what it does.  I think it could be made cleaner by renaming the method to `countValidationErrorsOfType` and pull the assertion out of it to the level where it is called.

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -390,18 +436,30 @@ private File exportGtfs(String namespace, DataSource dataSource, boolean fromEdi
...
      * database.
      */
     private void assertThatImportedGtfsMeetsExpectations(
-        String dbConnectionUrl,
+        Connection connection,
         String namespace,
         PersistenceExpectation[] persistenceExpectations
     ) throws SQLException {
+        // Store field mismatches here (to provide assertion statements with more details).
+        Multimap<String, ValuePair> fieldsWithMismatches = ArrayListMultimap.create();
```
evansiroky: I don't think it is necessary to include this `fieldsWithMismatches` variable.  Each PersistenceExpectation iterates through all found results and if the result is not found, that causes an error.  Then for each record this code asserts the expected vs found for each field.  I don't think this provides much valuable information for telling how every record is not like the one record the test is hoping to find.  The current tests do have a small amount of records, but if a test with a much larger dataset were added, I'm not sure how it'd be useful to describe how each record is different from the expected result.  I think it's best to just say, the record in this exact form was not found.

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -557,110 +659,33 @@ private boolean equalsWithNumericDelta(String val, RecordExpectation recordExpec
...
-            switch (expectedFieldType) {
-                case DOUBLE:
-                    return String.valueOf(doubleExpectation);
-                case INT:
-                    return String.valueOf(intExpectation);
-                case STRING:
-                    return stringExpectation;
-                default:
-                    return null;
+        if (mismatches != null) {
```
evansiroky: I don't think is necessary and request removal.  See discussion in [previous comment](https://github.com/conveyal/gtfs-lib/pull/143/files#r236448259).

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -140,6 +144,171 @@ public void canPreventSQLInjection() throws IOException, SQLException, InvalidNa
...
+                true
+        );
+        LOG.info("update {} output:", fareTable.name);
+        LOG.info(updateOutput);
+
+        FareDTO updatedFareDTO = mapper.readValue(updateOutput, fareDTOClass);
+
+        // make sure saved data matches expected data
+        assertThat(updatedFareDTO.fare_id, equalTo(updatedFareId));
+        assertThat(updatedFareDTO.fare_rules[0].fare_id, equalTo(updatedFareId));
```
evansiroky: We should also be testing that the value was updated in the database in these tests.

evansiroky: Referencing https://github.com/conveyal/gtfs-lib/issues/147

src/test/java/com/conveyal/gtfs/GTFSTest.java
landonreed: Why would we be looking to run this integration test with a very large dataset? I think I have to disagree on this one. Why wouldn't we want to improve the log output for a given test? If the integration test fails the PersistenceExpectation, we're leaving the maintainer with the task of combing through the expectations and the contents of the database to figure out what went wrong. It's been a couple of weeks, but I think this is exactly what happened to me while trying to add a new test.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
landonreed: Resolving this now that the ticket has been created.

src/test/java/com/conveyal/gtfs/GTFSTest.java
landonreed: Continue discussion in https://github.com/conveyal/gtfs-lib/pull/143#discussion_r242725670

src/test/java/com/conveyal/gtfs/GTFSTest.java
evansiroky: Ok, well yeah hopefully we can try to keep the test datasets as small as possible to make it easier to isolate issues.

src/test/java/com/conveyal/gtfs/GTFSTest.java
evansiroky: Was this going to be refactored?

evansiroky:
Assigning back to you @landonreed regarding https://github.com/conveyal/gtfs-lib/pull/143#discussion_r243451900.

src/test/java/com/conveyal/gtfs/GTFSTest.java
landonreed: Yes, apologies for missing that.

landonreed:
Thanks for catching that, @evansiroky. I addressed that comment and fixed some merge conflicts. Let me know if it looks good.

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -302,18 +380,34 @@ private File exportGtfs(String namespace, DataSource dataSource, boolean fromEdi
...
-        String dbConnectionUrl,
+        Connection connection,
         String namespace,
         PersistenceExpectation[] persistenceExpectations
     ) throws SQLException {
+        // Store field mismatches here (to provide assertion statements with more details).
+        Multimap<String, ValuePair> fieldsWithMismatches = ArrayListMultimap.create();
+        // Check that no validators failed during validation.
+        assertThat(
+            "No validators failed during GTFS import.",
```
evansiroky: This string should be reworded as it is a failure message to display in the event that the assertion fails.

PR #144, opened 2018-10-01
Graphql update

evansiroky:
This PR is in response to issue #125. Its objective is to simplify and reduce the number of SQL queries generated to handle a single GraphQL query (i.e. this is an optimization which operates within a single GraphQL query; it does not cache or optimize across multiple GraphQL queries).

- Update graphql-java from version 4.2 to 10.0
- Refactor NestedJDBCFetcher to make queries with joins instead of numerous individual queries
- Add dataloader integration to batch queries and cache sql queries
- Refactor initialization of graphQL executions

PR #149, opened 2018-11-20
Update GraphQL a little bit

evansiroky:
Take some very minor and reasonable changes that were included in #144 and call it good with updating graphQL a bit.

* Update GraphQL library from v4.2 to v11.0
* Add limit argument to routes nestedJdbcFetcher of stops
* Rename a few variables in JdbcFetcher so the query building makes more sense
* Add tests for highly nested GraphQL queries

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -353,7 +353,7 @@
     <dependency>
       <groupId>com.graphql-java</groupId>
       <artifactId>graphql-java</artifactId>
-      <version>4.2</version>
+      <version>11.0</version>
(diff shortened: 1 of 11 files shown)


abyrd:
My sense is that the super-nested query could be good for ensuring that the mechanism works properly - if it's a test of stability and functionality then it makes sense. I don't think it should be timed or have its performance otherwise measured though, as that could lead us to make compromises or time investments in development that do not improve our actual use cases.

PR #150, opened 2018-11-22
Next release (4.1.0)

evansiroky:
Includes a bunch of fixes and adds a bunch of tests

* allow null values for foreign ref fields allowed to be empty (#224)
* add delete feed feature to SQL code path
* more info in FeedLoadResult and TableLoadResult
* round outputted double feeds to avoid excessive precision
* properly store calendar_dates as schedule_exceptions
* make sure TimeFields are always exported in proper format
* ensure GTFS feeds with only calendar_dates are exported properly
* permit empty values if required field allows it 
* check for null namespace in tableExists method


landonreed:
Awesome, thanks for the summary @evansiroky. This is definitely a good time for a release.

PR #152, opened 2018-11-27
Rename variable to avoid confusion

evansiroky:
At first, I thought the variables `firstCalendarDate` and `lastCalendarDate` in the `ValidationResult` class were the first calendar date from the calendar_dates.txt file, but then I realized it was actually a result of calculating actual service of various days.  Renaming this variable should reflect this distinction and avoid future confusion.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/validator/ServiceValidator.java
+++ b/src/main/java/com/conveyal/gtfs/validator/ServiceValidator.java
@@ -185,8 +185,8 @@ public class ServiceValidator extends TripValidator {
                 if (date.isAfter(lastDate)) lastDate = date;
             }
             // Copy some useful information into the ValidationResult object to return to the caller.
-            validationResult.firstCalendarDate = firstDate;
-            validationResult.lastCalendarDate = lastDate;
+            validationResult.firstDateOfService = firstDate;
+            validationResult.lastDateOfService = lastDate;
             // Is this any different? firstDate.until(lastDate, ChronoUnit.DAYS);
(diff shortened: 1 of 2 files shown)


landonreed:
Wouldn't this be a breaking change? Is this variable change worth the effort to patch existing databases? I wonder if an explanatory comment would suffice instead of the variable name change.

landonreed:
This may be a dependent library, but the variable name is used both in datatools-server and datatools-ui. In short, this change has downstream effects and it can't be made without accompanying PRs and a script to migrate the `FeedVersion#validationResult` objects stored in MongoDB.

The variable name may not be as good as your proposal, but I don't think it warrants this level of effort to simply change the name. @abyrd, do you have any thoughts here?

evansiroky:
aw crap, didn't think of a database migration or its use in a js application or other language that isn't statically typed.

landonreed:
Yea, unfortunately it's a pretty involved process once we start using it in production environments.

abyrd:
I agree that @evansiroky 's variable names are an improvement, but yes, if this has repercussions in a database and JS applications we should just add comments instead of changing the names.

landonreed:
@evansiroky, are you OK with this conclusion? If so, could you revert the name change and simply add comments for now?

evansiroky:
Variable renamed back to original and comments added.

landonreed:
@evansiroky, I don't think the variable names were fixed in `ValidationResult.java`.

evansiroky:
Fixed

PR #153, opened 2018-11-27
Remove small table indexing

### LINKED ISSUE #132: SQL load fails if agency file is missing agency_id field
This is due to the attempt to create an index for the agency table, which is unlikely to ever actually improve query speed given the very small number of agencies in most GTFS feeds (1 or 2 usually).

landonreed:
This removes indexing for a couple of small tables, namely `agency` and `feed_info`.

Fixes #132.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/Table.java
+++ b/src/main/java/com/conveyal/gtfs/loader/Table.java
@@ -695,6 +695,10 @@ public class Table {
      * FIXME: add foreign reference indexes?
      */
     public void createIndexes(Connection connection, String namespace) throws SQLException {
+        if ("agency".equals(name)) {
+            LOG.info("Skipping indexes for {} table", name);
+            return;
+        }
         LOG.info("Indexing...");
         String tableName;
         if (namespace == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -695,6 +695,10 @@ public boolean isRequired () {
      * FIXME: add foreign reference indexes?
      */
     public void createIndexes(Connection connection, String namespace) throws SQLException {
+        if ("agency".equals(name)) {
+            LOG.info("Skipping indexes for {} table", name);
```
evansiroky: This probably doesn't need a template string.

landonreed: I used a template string in case we need to add tables to skip in the future and so that the agency table didn't get hard coded into the log message. Is that a strong enough reason? I realized I could also add a comment here explaining why it's being skipped, which I will do.

abyrd: You mean the template string in the log message? Personally I think it's fine. The overhead from substitution is trivial - the fact of logging at all probably has more impact, and as this is not a hotspot there is no need for optimization. Some people would always use template substitution in log messages for consistency - I often do it.

abyrd: Oh I see, there's only one possible value of `name` because of the conditional. This does make the template seem out of place. Is it expected that there will be more unindexed tables in the future?

landonreed: That was my intention with the use of the template string. I've just added feed_info to the conditional because there is absolutely no reason to index that table (it just wasn't causing problems like the agency index was).

PR #156, opened 2018-11-30
Update to version 4.1.x of maven-semantic-release

evansiroky:
maven-semantic-release v4.1.x+ now includes a verify-conditions step that checks if the maven project seems to be setup to be able to release to maven central.


landonreed:
LGTM.

PR #162, opened 2018-12-11
Fix duplicated schedule exceptions issue with multimap

### LINKED ISSUE #161: Auto-creation of schedule exceptions results in duplicate exception entries
For feeds with multiple removals/additions of service per date, the schedule exception auto creation process yields multiple exception entries during the feed snapshotting process. This is due to the use of a List data structure rather than one that contains unique entries:

https://github.com/conveyal/gtfs-lib/blob/5b87be8b425480f62ce9a5cda753a4626bc7333d/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsSnapshotter.java#L218

landonreed:
This fixes and simplifies the data structures used in the snapshotting process to create schedule exceptions from calendar dates. Fixes #161.


landonreed:
Merging for now because I need the snapshot on https://oss.sonatype.org/content/repositories/snapshots/com/conveyal/gtfs-lib/

evansiroky:
Good catch.

Issue #109, opened 2018-04-05
Updates to "exemplar" fields not applied to linked fields


landonreed:
This issue was uncovered when a user reported that updates to the `wheelchair_accessible` field on routes were being silently dropped.  For one, the field was not present in the Table definition.  The second more critical issue discovered is that this "exemplar" field and fields on other tables (patterns and pattern stops) are not being applied to linked fields in referring tables.  As a concrete example: the `trips#wheelchair_accessible` field is derived from the `route#wheelchair_accessible` field in the editor.  Another example: `pattern_stops` has a number of fields (e.g., `pickup_type` and `timepoint`) that need to be applied to all stop_times that refer to trips on this pattern.

PR #164, opened 2018-12-18
Update trip#shape_id and frequency stop_times on pattern updates

### LINKED ISSUE #165: Editor: pattern stop travel time updates not applied to frequency trip arrival/departure times
In the editor, a pattern that uses frequencies will have referencing trips with stop times that adhere to the default travel and dwell times values assigned to each of the pattern's stops. This is true when new trips/frequencies are added to an existing "use-frequency" pattern. Currently, however,  if these default travel or dwell times were to change, the referencing stop_times would not have their values updated.

### LINKED ISSUE #173: Editor: Frequency entries can be added to patterns that do not "use frequencies"
I discovered while building tests for `JdbcTableWriter` that a trip with a frequency entry can be added to a pattern that does not have useFrequency set to `true`.

landonreed:
In addition to some formatting tweaks and added comments, this seeks to address the underlying cause for conveyal/datatools-ui#385. It is another example of updates for "exemplar" fields not making their way back to the referencing entities (#109).

It also fixes #165, where stop_time arrivals and departures were not kept in sync with updates to the underlying pattern travel times. This is only a concern with frequency-based patterns, but the same method could potentially be utilized for updating travel times in timetable-based trips if needed in the future.

Finally, it adds some tests to `JdbcTableWriter` for creating frequency trips and fixes #173, which was discovered in the process of writing these tests.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
+++ b/src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
@@ -77,9 +77,11 @@ public class JdbcTableWriter implements TableWriter {
 
         // TODO: verify specTable.name is ok to use for constructing dynamic sql statements
         this.specTable = specTable;
+        // Below is a bit messy because the connection field on this class is set to final and we cannot set this to
+        // the optionally passed-in connection with the ternary statement unless connection1 already exists.
         Connection connection1;
         try {
-            connection1 = dataSource.getConnection();
+            connection1 = this.dataSource.getConnection();
         } catch (SQLException e) {
             e.printStackTrace();
             connection1 = null;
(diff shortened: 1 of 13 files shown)


src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
```diff
@@ -492,6 +570,41 @@ private void updateChildTable(ArrayNode subEntities, Integer id, boolean isCreat
...
+     * @throws SQLException
+     */
+    private int updateStopTimesForPatternStop(ObjectNode patternStop, int previousTravelTime) throws SQLException {
+        String sql = String.format(
+            "update %s.stop_times st set arrival_time = ?, departure_time = ? from %s.trips t " +
+                "where st.trip_id = t.trip_id AND t.%s = ? AND st.%s = ?",
+            tablePrefix,
+            tablePrefix,
+            "pattern_id",
+            "stop_sequence"
```
evansiroky: Since these are hardcoded strings, they should be included in the base string.

src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
```diff
@@ -492,6 +570,41 @@ private void updateChildTable(ArrayNode subEntities, Integer id, boolean isCreat
...
+        PreparedStatement statement = connection.prepareStatement(sql);
+        int oneBasedIndex = 1;
+        int travelTime = patternStop.get("default_travel_time").asInt();
+        int arrivalTime = previousTravelTime + travelTime;
+        statement.setInt(oneBasedIndex++, arrivalTime);
+        int dwellTime = patternStop.get("default_dwell_time").asInt();
+        statement.setInt(oneBasedIndex++, arrivalTime + dwellTime);
+        // Set "where clause" with value for pattern_id and stop_sequence
+        statement.setString(oneBasedIndex++, patternStop.get("pattern_id").asText());
+        statement.setInt(oneBasedIndex++, patternStop.get("stop_sequence").asInt());
```
evansiroky: Using a variable for the prepared statement variable index seems a little too fancy.  Since the variables aren't dynamically created and the length of the parameters isn't varying, I think it may be best to simply hardcode the field position number.

landonreed: It seems more maintainable and less error-prone to me to use an incrementing variable. As long as the you're setting the parameters in the correct order, adding a new parameter is simply a matter of inserting it in the right spot and using the same `oneBasedIndex`. If we were to use hardcoded numbers, updating the statement could involve updating multiple numbers one by one if the new parameter is placed anywhere except the end.

evansiroky: ok, I don't have a super strong opinion on this, so go ahead.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -307,6 +316,169 @@ public void canCreateUpdateAndDeleteRoutes() throws IOException, SQLException, I
...
+        JdbcTableWriter createTripWriter = createTestTableWriter(Table.TRIPS);
+        createTripWriter.create(mapper.writeValueAsString(tripInput), true);
+    }
+
+    /**
+     * Checks that creating a frequency trip functions properly. This also updates the pattern to include pattern stops,
+     * which is a prerequisite for creating a frequency trip with stop times.
+     */
+    @Test
+    public void canCreateFrequencyForFrequencyPattern() throws IOException, SQLException, InvalidNamespaceException {
```
evansiroky: Great start!  This should be renamed to indicate that it also does some checking of the updating and deleting of some trips.

Issue #167, opened 2018-12-19
Validator: parity with Google transitfeed feedvalidator


landonreed:
I think it would be useful to review the [list of validation errors and warnings](https://github.com/google/transitfeed/wiki/FeedValidatorErrorsAndWarnings) thrown by the Google transitfeed validator to ensure we're not missing anything glaring in gtfs-lib. I'm not sure this has been done on the (at least on the RDBMS code path) and I would like to say with confidence that gtfs-lib is performing every check made by this peer tool.

PR #169, opened 2018-12-19
Validate stop names and trip headsigns

landonreed:
This adds validation checks for:
- stop names (that they exist and are not the same as the stop description) and
- trip headsigns (that they do not contain route name and do not begin with "to" or "towards" per http://gtfs.org/best-practices#tripstxt)

References #167.


landonreed:
Good point. This goes for a lot of other validators, too. I just created #174 to track this.

Issue #186, opened 2019-01-22
GraphQL schema route#route_type field type is string


landonreed:
In the GraphQL schema definition, `route#route_type` is set to be a string type, not an int (which is what it should be).

https://github.com/conveyal/gtfs-lib/blob/040a9f8b7317e48a5091a4717b23b215d828f302/src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java#L266

PR #188, opened 2019-01-23
Fix pattern#direction_id import and route_type field type

### LINKED ISSUE #187: pattern#direction_id not imported from exemplar trip in RDBMS code path
When loading a GTFS feed into PostgreSQL, the `trip#direction_id` field is not used to populate `pattern#direction_id`.

landonreed:
Fixes #187 and #186.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
 
   <groupId>com.conveyal</groupId>
   <artifactId>gtfs-lib</artifactId>
-  <version>4.8</version>
+  <version>4.2.5-SNAPSHOT</version>
   <packaging>jar</packaging>
 
   <licenses>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -8,7 +8,7 @@
 
   <groupId>com.conveyal</groupId>
   <artifactId>gtfs-lib</artifactId>
-  <version>4.2.5-SNAPSHOT</version>
+  <version>4.8</version>
```
evansiroky: ???

landonreed: This shouldn't have been checked in, sorry about that. In order to use a development copy of gtfs-lib with datatools-server, I manually change the version in the pom to a non-existent future version to match datatools-server so that Intellij will use the local code rather than a jar from the maven repo.

Issue #146, opened 2018-11-05
'Get all shapes' GraphQL query (for rendering lines on map) breaks for large feeds


landonreed:
Currently in datatools-ui, there is a GraphQL request that is made to show all of a feed's patterns on a map. This breaks down for very large GTFS feeds (like the NL) and becomes slow (5-10 seconds) for large-ish feeds like NYC buses. This problem was partially the inspiration for #125; however, that issue and its PR ended up adding too much complexity in pursuit of too little (or none at all) optimization.

@evansiroky has done some initial benchmarking to isolate some of this problem, but there may need to be some more tests to isolate this problem. The solution is undetermined, but some suggestions have centered on storing some sort of cached or simplified representation of the shapes rather than fetching them all from a GraphQL query for patterns (with nested shapes).

landonreed:
Rather than using GraphQL for this, I wonder if it might be more effective to just generate vector tiles for all shapes as part of the feed processing step with something like https://github.com/ElectronicChartCentre/java-vector-tile. This might be more appropriate in https://github.com/conveyal/datatools-server, but if other projects using gtfs-lib need this perhaps the code could go here.

PR #190, opened 2019-01-31
Copy shape on update

landonreed:
When multiple patterns share a reference to the same shape (for example, patterns for BART routes that travel along the same rail corridor but have short and long runs), editing the shape for one of these patterns can cause undesired changes for the other patterns. Because we don't really have a good way to handle shared shapes in the editor, we need an update to a shared shape to generate a new shape_id and leave the previous shape untouched. This PR handles this logic and also changed the log level for some noisy SQL logs to debug because they may be contributing to some performance issues for large SQL queries (see #146).


src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -366,12 +382,48 @@ private static PatternDTO createSimpleRouteAndPattern(String routeId, String pat
...
+        // Verify that shape_id is shared.
+        assertThat(patternWithSharedShape.shape_id, equalTo(sharedShapeId));
+        // Update any field on one of the patterns.
+        JdbcTableWriter patternUpdater = createTestTableWriter(Table.PATTERNS);
+        patternWithSharedShape.name = "The shape_id should update";
+        String sharedPatternOutput = patternUpdater.update(patternWithSharedShape.id, mapper.writeValueAsString(patternWithSharedShape), true);
+        // The output should contain a new backend-generated shape_id.
+        PatternDTO updatedSharedPattern = mapper.readValue(sharedPatternOutput, PatternDTO.class);
+        LOG.info("Updated pattern output: {}", sharedPatternOutput);
+        assertThat(updatedSharedPattern.shape_id, not(equalTo(sharedShapeId)));
```
evansiroky: We need to get into the habit of also checking that the database looks good in these tests.  Funky stuff could happen in the database that aren't reflected in the output from gtfs-lib.  Please add another few lines of code that make sure that the `shape_id` of the pattern with the secondPatternId doesn't equal the previous shared shape id.

landonreed:
Thanks for the review, @evansiroky. I've updated the test to check that the updated shape_id value persists in the database.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -421,7 +421,15 @@ public void shouldChangeShapeIdOnPatternUpdate() throws IOException, SQLExceptio
         // The output should contain a new backend-generated shape_id.
         PatternDTO updatedSharedPattern = mapper.readValue(sharedPatternOutput, PatternDTO.class);
         LOG.info("Updated pattern output: {}", sharedPatternOutput);
-        assertThat(updatedSharedPattern.shape_id, not(equalTo(sharedShapeId)));
+        String newShapeId = updatedSharedPattern.shape_id;
+        assertThat(newShapeId, not(equalTo(sharedShapeId)));
+        // Ensure that pattern record in database reflects updated shape ID.
+        assertThatSqlQueryYieldsRowCount(String.format(
+            "select * from %s.%s where shape_id='%s'",
```
evansiroky: I think we're almost there, but this assertion would pass if somehow the pattern with `patternId` was somehow updated to have the value of `newShapeId`.  This could be improved by adding a where clause for the pattern id so that a more exact expectation of data integrity can happen.

landonreed: Are you suggesting that this where clause be changed to `where shape_id=$newShapeId AND pattern_id=$secondPatternId`?

evansiroky: Yes, exactly.

abyrd: Great to see very specific assertions happening.

Issue #170, opened 2018-12-31
JdbcGtfsExporter: cleanUpZipFile() fails on Windows


simonsouth:
On Windows, `JdbcGtfsExporter.cleanUpZipFile()` fails with an `IllegalArgumentException` as the method it uses for generating a URI works only on systems with UNIX-like file paths.

With [Data Tools](https://github.com/conveyal/datatools-server) deployed on Windows, this means exporting a GTFS feed partially fails with a log message like the following:

```
Dec 30 18:05:41.308 [pool-1-thread-2] ERROR c.c.gtfs.loader.JdbcGtfsExporter:310 - Exception while creating snapshot: java.lang.IllegalArgumentException: Illegal character in opaque part at index 13: jar:file://C:\Users\simon\AppData\Local\Temp\snapshot3620290097384378083zip
java.lang.IllegalArgumentException: Illegal character in opaque part at index 13: jar:file://C:\Users\simon\AppData\Local\Temp\snapshot3620290097384378083zip
        at java.net.URI.create(URI.java:852)
        at com.conveyal.gtfs.loader.JdbcGtfsExporter.cleanUpZipFile(JdbcGtfsExporter.java:328)
        at com.conveyal.gtfs.loader.JdbcGtfsExporter.exportTables(JdbcGtfsExporter.java:301)
        at com.conveyal.datatools.editor.jobs.ExportSnapshotToGTFSJob.jobLogic(ExportSnapshotToGTFSJob.java:51)
        at com.conveyal.datatools.common.status.MonitorableJob.run(MonitorableJob.java:126)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.net.URISyntaxException: Illegal character in opaque part at index 13: jar:file://C:\Users\simon\AppData\Local\Temp\snapshot3620290097384378083zip
        at java.net.URI$Parser.fail(URI.java:2848)
        at java.net.URI$Parser.checkChars(URI.java:3021)
        at java.net.URI$Parser.parse(URI.java:3058)
        at java.net.URI.<init>(URI.java:588)
        at java.net.URI.create(URI.java:850)
        ... 7 more
[...]

PR #192, opened 2019-02-11
Bug Fix Release

### LINKED ISSUE #189: Copy on update for shapes referenced by multiple patterns
>I found an issue with editing certain segments on routes that have a shape that extends past the final stop of a pattern.  When editing on a segment between the final 2 stops of a route, any changes would also apply to the remaining shape beyond the final stop of the pattern as well.  Perhaps the right thing to do here would be to clone the shape and save a new version of it.  Or maybe there's a way to have a shape edit apply to multiple patterns?  Not sure here.

_Originally posted by @evansiroky in https://github.com/conveyal/datatools-ui/pull/401#issuecomment-458434242_

landonreed:
~Once #190 is merged~ (it is merged!), I would like to create a new bug fix release.

### Issues fixed
#192 
#189
#170
#186
#187


evansiroky:
Fix for #193 is here: #196.

PR #200, opened 2019-02-15
Tweaks for MTC enhancements

landonreed:
This PR modifies and makes static/public a few internal methods for `Table` and `JdbcTableLoader` that are used in https://github.com/conveyal/datatools-server/pull/186.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
+++ b/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
@@ -209,7 +209,7 @@ public class JdbcGtfsLoader {
         // FIXME is this extra CSV reader used anymore? Check comment below.
         // First, inspect feed_info.txt to extract the ID and version.
         // We could get this with SQL after loading, but feed_info, feed_id and feed_version are all optional.
-        CsvReader csvReader = getCsvReader(Table.FEED_INFO);
+        CsvReader csvReader = Table.FEED_INFO.getCsvReader(zip, errorStorage);
         String feedId = "", feedVersion = "";
         if (csvReader != null) {
             // feed_info.txt has been found and opened.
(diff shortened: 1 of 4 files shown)


src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -259,27 +258,30 @@ private void registerFeed (File gtfsFile) {
     /**
      * In GTFS feeds, all files are supposed to be in the root of the zip file, but feed producers often put them
      * in a subdirectory. This function will search subdirectories if the entry is not found in the root.
-     * It records an error if the entry is in a subdirectory.
+     * It records an error if the entry is in a subdirectory (as long as errorStorage is not null).
      * It then creates a CSV reader for that table if it's found.
+     *
+     * This method is static so that it can be reused outside of JdbcGtfsLoader for getting CSV readers for zip files
+     * outside of the load procedure.
```
evansiroky: This comment makes me wonder if it may be more appropriate to have this method be within the Table class instead. Is that possible?

src/main/java/com/conveyal/gtfs/loader/ReferenceTracker.java
```diff
@@ -0,0 +1,9 @@
+package com.conveyal.gtfs.loader;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class ReferenceTracker {
```
evansiroky: Please add a comment describing what the purpose of this class is.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -485,6 +476,53 @@ private int loadInternal (Table table) throws Exception {
         return numberOfRecordsLoaded;
     }
 
+    /**
+     * Returns the index of the specified field name for the array of fields provided.
+     * @param fields array of fields (intended to be derived from the headers of a csv text file)
+     * @param fieldName string name of the field to find
+     */
+    public static int getFieldIndex(Field[] fields, String fieldName) {
```
evansiroky: Yeah, seeing how these are called from https://github.com/conveyal/datatools-server/pull/186/files#diff-b601a83ada6de294158d19894ceefe57, I think it might make more sense to put these methods in the Table class.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -485,6 +476,53 @@ private int loadInternal (Table table) throws Exception {
...
+        }
+        return fieldIndex;
+    }
+
+    /**
+     * Returns the index of the key field within the array of fields provided for a given table.
+     * @param table table from which to derive the key field
+     * @param fields array of fields (intended to be derived from the headers of a csv text file)
+     */
+    public static int getKeyFieldIndex(Table table, Field[] fields) {
```
evansiroky: Candidate for moving into Table class.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -485,6 +476,53 @@ private int loadInternal (Table table) throws Exception {
...
+        String keyField = table.getKeyFieldName();
+        return getFieldIndex(fields, keyField);
+    }
+
+    /**
+     * For an array of field headers, returns the matching set of {@link Field}s for a {@link Table}. If errorStorage is
+     * not null, errors related to unexpected or duplicate header names will be stored.
+     */
+    public static Field[] getFieldsFromFieldHeaders(String[] headers, Table table, SQLErrorStorage errorStorage) {
```
evansiroky: Candidate for moving into Table class.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -562,24 +600,12 @@ private void setFieldToNull(boolean postgresText, String[] transformedStrings, i
      *
      * TODO add a test including SQL injection text (quote and semicolon)
      */
-    public String sanitize (String string) throws SQLException {
+    public static String sanitize (String string, SQLErrorStorage errorStorage) {
         String clean = string.replaceAll("[^\\p{Alnum}_]", "");
```
evansiroky: I don't quite understand this regex. Can you make a comment for it?

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -870,75 +893,86 @@ public Table getParentTable() {
         return parentTable;
     }
 
+    public Set<NewGTFSError> checkReferencesAndUniqueness(String keyValue, int lineNumber, Field field, String value, ReferenceTracker referenceTracker) {
```
evansiroky: Needs javadoc

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
evansiroky: Actually, this one doesn't have a Table argument, so maybe this should get put into the Field class.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
landonreed: @abyrd, can you provide a description of this? (I can add your comments to code).

landonreed:
@evansiroky, thanks for the suggestions. I've addressed the comment gaps and moved methods to Table/Field. I don't know exactly what the regex does, but I've tagged @abyrd who wrote that method originally. If you'd like we can wait for his comments before merge.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
abyrd: Sorry for the undocumented regex, obviously these things should always have a comment. This removes all characters that are not alphanumeric or an underscore. The reasoning being that all GTFS field names are composed of only alphanumeric characters and underscores, and limiting all supplied strings to those characters should completely preclude SQL injection.

PR #201, opened 2019-02-19
Refactor field validation in order to store/validate bad values properly

### LINKED ISSUE #191: Check GTFS fields for new lines and other special characters
According to the [GTFS spec](https://developers.google.com/transit/gtfs/reference/#file_requirements),

> Field values may not contain tabs, carriage returns or new lines.

I don't believe we are checking for this anywhere and recently a GTFS feed for Tompkins Consolidated Area Transit (TCAT) passed our validator and ended up choking OpenTripPlanner's OBA GTFS reader.

### LINKED ISSUE #202: Special handling for calendar_date#service_id reference not found in calendars.txt
As part of the MTC enhancements added for supporting merging feeds, a foreign reference was added to the calendar_dates#service_id table. Doing so allows us to scope the service_id and possibly remap it if conflicts are encountered between two feeds.

https://github.com/conveyal/gtfs-lib/blob/5ec52854ac2c39dd94f8ac9d0f0fc051cf53d312/src/main/java/com/conveyal/gtfs/loader/Table.java#L138-L142

However, this change will have the effect of causing false positives for referential integrity errors encountered during feed loading for GTFS feeds like TriMet's, which define service dates using calendar_dates (exception type = 1) alone. I have a fix inbound for this issue, which will only store the ref integrity error if the exception type is not `1`.

landonreed:
This PR fixes #191 by modifying how values encountered during GTFS loads are validated and cleaned. It uses a helper return class `ValidateFieldResult` to store both the cleaned value and any errors found during validation, which then are bubbled up and handled in `JdbcGtfsLoader`.

It also adds a new error type `ILLEGAL_FIELD_VALUE` to resolve the primary issue of not tracking instances where bad character sequences where found in a GTFS feed.

This also fixes #202.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/DateListField.java
+++ b/src/main/java/com/conveyal/gtfs/loader/DateListField.java
@@ -1,5 +1,6 @@
 package com.conveyal.gtfs.loader;
 
+import com.conveyal.gtfs.error.NewGTFSError;
 import com.conveyal.gtfs.storage.StorageException;
 
 import java.sql.Array;
@@ -7,9 +8,8 @@ import java.sql.JDBCType;
 import java.sql.PreparedStatement;
 import java.sql.SQLType;
 import java.util.Arrays;
-import java.util.stream.Collectors;
-
-import static com.conveyal.gtfs.loader.DateField.validate;
+import java.util.Collections;
+import java.util.Set;
 
 public class DateListField extends Field {
 
@@ -18,19 +18,22 @@ public class DateListField extends Field {
     }
 
     @Override
-    public String validateAndConvert(String original) {
-        // FIXME
+    public ValidateFieldResult<String> validateAndConvert(String original) {
+        // FIXME: This function is currently only used during feed loading. Because the DateListField only exists for the
+        //  schedule exceptions table (which is not a GTFS spec table), we do not expect to ever call this function on
+        //  this table/field.
         return null;
     }
 
     @Override
-    public void setParameter(PreparedStatement preparedStatement, int oneBasedIndex, String string) {
+    public Set<NewGTFSError> setParameter(PreparedStatement preparedStatement, int oneBasedIndex, String string) {
         try {
             String[] dateStrings = Arrays.stream(string.split(","))
                     .map(DateField::validate)
                     .toArray(String[]::new);
             Array array = preparedStatement.getConnection().createArrayOf("text", dateStrings);
             preparedStatement.setArray(oneBasedIndex, array);
+            return Collections.EMPTY_SET;
         } catch (Exception ex) {
             throw new StorageException(ex);
         }
--- a/src/main/java/com/conveyal/gtfs/loader/DateListField.java
+++ b/src/main/java/com/conveyal/gtfs/loader/DateListField.java
@@ -11,6 +11,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Set;
 
+/**
+ * Field type for a list of date strings. This type is ONLY used for the editor-specific schedule exceptions table, so
+ * not all methods are supported (e.g., validateAndConvert has no functionality because it is only called on GTFS tables).
+ */
 public class DateListField extends Field {
 
     public DateListField(String name, Requirement requirement) {
@@ -19,10 +23,11 @@ public class DateListField extends Field {
 
     @Override
     public ValidateFieldResult<String> validateAndConvert(String original) {
-        // FIXME: This function is currently only used during feed loading. Because the DateListField only exists for the
-        //  schedule exceptions table (which is not a GTFS spec table), we do not expect to ever call this function on
-        //  this table/field.
-        return null;
+        // This function is currently only used during feed loading. Because the DateListField only exists for the
+        // schedule exceptions table (which is not a GTFS spec table), we do not expect to ever call this function on
+        // this table/field.
+        // TODO: is there any reason we may want to add support for validation?
+        throw new UnsupportedOperationException("Cannot call validateAndConvert on field of type DateListField because this is not a supported GTFS field type.");
     }
 
     @Override
--- a/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
+++ b/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
@@ -386,7 +386,7 @@ public class JdbcGtfsLoader {
                 String string = csvReader.get(f);
                 // Use spec table to check that references are valid and IDs are unique.
                 Set<NewGTFSError> errors = table.checkReferencesAndUniqueness(keyValue, lineNumber, field, string, referenceTracker);
-                // Check for special case with calendar_dates where addedd service should not trigger ref. integrity
+                // Check for special case with calendar_dates where added service should not trigger ref. integrity
                 // error.
                 if (
                     table.name.equals("calendar_dates") &&
--- a/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
+++ b/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
@@ -1,6 +1,7 @@
 package com.conveyal.gtfs.loader;
 
 import com.conveyal.gtfs.error.NewGTFSError;
+import com.conveyal.gtfs.error.NewGTFSErrorType;
 import com.conveyal.gtfs.error.SQLErrorStorage;
 import com.conveyal.gtfs.storage.StorageException;
 import com.csvreader.CsvReader;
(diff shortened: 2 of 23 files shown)


src/main/java/com/conveyal/gtfs/loader/DateListField.java
```diff
@@ -18,19 +18,22 @@ public DateListField(String name, Requirement requirement) {
     }
 
     @Override
-    public String validateAndConvert(String original) {
-        // FIXME
+    public ValidateFieldResult<String> validateAndConvert(String original) {
+        // FIXME: This function is currently only used during feed loading. Because the DateListField only exists for the
```
evansiroky: I don't like seeing FIXMEs in comments. Can we just remove the FIXME from the comment and call it good for now? If not, how would we go about fixing this?

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -385,7 +385,17 @@ private int loadInternal (Table table) throws Exception {
...
                 Set<NewGTFSError> errors = table.checkReferencesAndUniqueness(keyValue, lineNumber, field, string, referenceTracker);
-                errorStorage.storeErrors(errors);
+                if (
+                    table.name.equals("calendar_dates") &&
+                    "1".equals(csvReader.get(Field.getFieldIndex(fields, "exception_type")))
+                ){
+                    // Do not record bad reference errors for calendar date entries that add service
+                    // (exception type=1) because a corresponding service_id in calendars.txt is not required in
+                    // this case.
+                    LOG.info("Skipping bad reference for calendar_date adding service (service_id={}).", keyValue);
```
evansiroky: Are we absolutely sure that no other errors are skipped here? Maybe to future-proof this, it'd be a good idea to explicitly check if the exact error we're looking for occurred and only then do this thing. Also, I'd reword the log statement to not say that it's skipping something bad, but maybe something like an added service calendar_date entry was found that doesn't need to have an entry in the calendar file.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -385,7 +385,17 @@ private int loadInternal (Table table) throws Exception {
...
+                }
+                // In all other cases (outside of the calendar_dates special case), store the reference errors found.
+                else errorStorage.storeErrors(errors);
```
evansiroky: Since the above if statement has braces, please include it with this else statement for readability.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -466,13 +476,27 @@ public void setValueForField(Table table, int fieldIndex, int lineNumber, Field
...
-                else field.setParameter(insertStatement, fieldIndex + 2, string);
+                Set<NewGTFSError> errors;
+                if (postgresText) {
+                    ValidateFieldResult<String> result = field.validateAndConvert(string);
+                    // If the result is null, use the null-setting method.
+                    if (result.clean == null) setFieldToNull(postgresText, transformedStrings, fieldIndex, field);
+                    // Otherwise, set the cleaned field according to its index.
+                    else transformedStrings[fieldIndex + 1] = result.clean;
+                    errors = result.errors;
+                } else errors = field.setParameter(insertStatement, fieldIndex + 2, string);
```
evansiroky: Please use braces here too.

src/main/java/com/conveyal/gtfs/loader/StringListField.java
```diff
@@ -14,17 +17,20 @@ public StringListField(String name, Requirement requirement) {
     }
 
     @Override
-    public String validateAndConvert(String original) {
-        // FIXME
+    public ValidateFieldResult<String> validateAndConvert(String original) {
+        // FIXME: This function is currently only used during feed loading. Because the StringListField only exists for the
```
evansiroky: Remove FIXME or fix the underlying issue, please.

src/test/java/com/conveyal/gtfs/loader/FieldTests.java
```diff
@@ -30,24 +27,54 @@ public void dateFieldParseTest() {
...
+            assertThat("Returned value matches the well-formed date.", result.clean.equals(goodDate));
         }
 
     }
+
+    /**
+     * Make sure {@link Field#cleanString(ValidateFieldResult)} catches and removes illegal character sequences.
+     */
+    @Test
+    public void illegalCharacterParseTest() {
```
evansiroky: Each of these tests should be refactored into[ parameritized tests](https://github.com/junit-team/junit4/wiki/parameterized-tests). This way is preferable because each test case will be ran regardless of whether a previous test case failed.

src/main/java/com/conveyal/gtfs/loader/DateListField.java
abyrd: Can you explain why you want fix-me comments to be removed? I have added many hundreds of such comments all over our code bases, so if there is a decision to remove them or not do this it will make a big difference for me. While working on something, various problems and concerns are identified, and it's a real loss for those to be forgotten. It's impossible to branch off development every time a new idea comes up, and it's often better to let ideas mature and confront them with later more complete stages of implementation. Often those ideas or problems need to be seen by the next person working on the code.

abyrd: The above is a general response about fix-me comments. In this specific case though, I agree that this should just be fixed. Where a function in an interface is not needed on a special-purpose subclass, in my opinion the proper thing to do is to throw an "operation not supported" exception, and include a comment much like this one that explains why the operation is not supported.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
abyrd: A convention we have followed for a few years is that braces may be omitted if the statement is on the same line as the conditional keyword. Sometimes that makes things easier to read. But generally I agree, I'd default to wrapping all conditional bodies in braces.

src/main/java/com/conveyal/gtfs/loader/StringListField.java
abyrd: Again if this concrete class does not need to implement a method defined on its interface or superclass, it should throw an unsupported operation exception and include a comment about why it doesn't need to support the operation.

src/test/java/com/conveyal/gtfs/loader/FieldTests.java
abyrd: Interesting and good to know about the parameterized test runner. It does seem like a more "correct" way to express running the same test N times with different input. It's unfortunate though that apparently the parameters are Objects instead of specific types, and I find the use of constructors to set fields on a class instance to be an awkward way to get the parameters into the test. But I suppose once you're used to seeing the pattern it's readable.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
abyrd: I agree on both points - good to check for a specific error type and make messages as clear as we can.

src/main/java/com/conveyal/gtfs/loader/DateListField.java
evansiroky: In the general sense, FIXME to me seems to indicate something is wrong and must be corrected. I'd prefer using TODO in those cases when an idea should be fleshed out a little better.

abyrd: Ah, if it's just a question of not exaggerating severity, as well as fixing small things now instead of later, then I see what you mean.

src/test/java/com/conveyal/gtfs/loader/FieldTests.java
landonreed: I find this construction very awkward as well, and had simply copied the test for the DateField type directly above for this new illegal characters test. If both of these tests need revision, I suggest we pull a new ticket and handle this test issue separately.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsLoader.java
```diff
@@ -385,17 +386,32 @@ private int loadInternal (Table table) throws Exception {
                 String string = csvReader.get(f);
                 // Use spec table to check that references are valid and IDs are unique.
                 Set<NewGTFSError> errors = table.checkReferencesAndUniqueness(keyValue, lineNumber, field, string, referenceTracker);
+                // Check for special case with calendar_dates where addedd service should not trigger ref. integrity
```
evansiroky: typo: "addedd"

PR #205, opened 2019-02-27
Allow nested queries where field names don't  match

landonreed:
The GraphQL nested queries up until this change relied on parent tables with foreign refs sharing
the same field name as the primary key on the child table. For example, trips could be joined to a
route on the shared field route_id. This change permits nesting children stops under a parent stop
using stops#parent_station, which will be joined on the parent join value found in stops#stop_id.

refs conveyal/datatools-ui#428


landonreed:
@evansiroky, good point. I just added a test for this (and updated a few snapshots that changed as a result).

PR #208, opened 2019-03-13
Fix overlapping trips in block check

### LINKED ISSUE #207: Overlapping trips in block check is disabled
There is an `OverlappingTripsValidator` but it is full of FIXMEs and never actually gets run:

https://github.com/conveyal/gtfs-lib/blob/53655768f6e5f624812a03be05d60c0f96c54f24/src/main/java/com/conveyal/gtfs/validator/NewTripTimesValidator.java#L44-L52

I think we need the `ServiceValidator#serviceInfoForServiceId` map in `OverlappingTripsValidator`, but I’m not sure the best way to share that between the two validators. We could assign it to the validation result, which gets passed to each trip validator on completion, but I don't know if this should be part of the validation result. I'll submit a PR with some work I've done here and some tests to help reproduce.

landonreed:
### Checklist

### Description

This PR fixes #207 by re-enabling the `OverlappingTripsValidator` check and adding a corresponding test.


landonreed:
@abyrd and @evansiroky, this is ready for review, and I would like to know if you have any specific thoughts regarding adding `serviceInfoForServiceId` to `ValidationResult`. This was done to give access to this information to the `OverlappingTripsValidator`, but I'm not sure if we should be storing that data in the validation result. Are there any other strategies we could use?

src/main/java/com/conveyal/gtfs/validator/OverlappingTripValidator.java
```diff
@@ -33,94 +35,66 @@ public OverlappingTripValidator(Feed feed, SQLErrorStorage errorStorage) {
...
     public void complete (ValidationResult validationResult) {
+        // Iterate over each block and determine if there are any trips that overlap one another.
         for (String blockId : blockIntervals.keySet()) {
             List<BlockInterval> intervals = blockIntervals.get(blockId);
-            Collections.sort(intervals, Comparator.comparingInt(i -> i.startTime));
-            int i2offset = 0;
+            intervals.sort(Comparator.comparingInt(i -> i.startTime));
+            // Iterate over each interval (except for the last) comparing it to every other interval (so the last interval
+            // is handled through the course of iteration).
             // FIXME this has complexity of n^2, there has to be a better way.
```
evansiroky: I think this FIXME comment could be removed. Also, this probably isn't a bottleneck since there are probably not millions of different intervals per block.

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -125,6 +126,25 @@ public void canLoadFeedWithBadDates () {
         );
     }
 
+    /**
+     * Tests that a GTFS feed with bad date values in calendars.txt and calendar_dates.txt can pass the integration test.
+     */
+    @Test
+    public void canLoadFeedWithOverlappingTrips () {
```
evansiroky: Love seeing more tests! 😍

PR #210, opened 2019-03-18
Normalize stop times travel times for set of pattern stops

landonreed:
### Checklist

### Description

Adds feature to normalize all stop times for a set of pattern stops in order to simplify updating
existing trips in bulk when a pattern changes (e.g., travel times change or new pattern stops are
inserted).


src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -484,6 +486,50 @@ public void canCreateUpdateAndDeleteFrequencyTripForFrequencyPattern() throws IO
...
+        final Table tripsTable = Table.TRIPS;
+        int initialTravelTime = 60; // one minute
+        int startTime = 6 * 60 * 60; // 6AM
+        String patternId = "123456";
+        PatternStopDTO[] patternStops = new PatternStopDTO[]{
+            new PatternStopDTO(patternId, firstStopId, 0),
+            new PatternStopDTO(patternId, lastStopId, 1)
+        };
+        patternStops[1].default_travel_time = initialTravelTime;
+        PatternDTO pattern = createRouteAndPattern(UUID.randomUUID().toString(), patternId, "Pattern A", null, new ShapePointDTO[]{}, patternStops, 0);
```
evansiroky: Please format this line so it isn't so long.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -484,6 +486,50 @@ public void canCreateUpdateAndDeleteFrequencyTripForFrequencyPattern() throws IO
...
+        // Create trip with travel times that match pattern stops.
+        TripDTO tripInput = constructTimetableTrip(pattern.pattern_id, pattern.route_id, startTime, initialTravelTime);
```
evansiroky: Please format this line so it isn't so long.

evansiroky: actually, this line looks ok in the IDE, disregard.

src/test/java/com/conveyal/gtfs/loader/JDBCTableWriterTest.java
```diff
@@ -484,6 +486,50 @@ public void canCreateUpdateAndDeleteFrequencyTripForFrequencyPattern() throws IO
...
+        JdbcTableWriter patternUpdater = createTestTableWriter(Table.PATTERNS);
+        int updatedTravelTime = 3600; // one hour
+        pattern.pattern_stops[1].default_travel_time = updatedTravelTime;
+        String updatedPatternOutput = patternUpdater.update(pattern.id, mapper.writeValueAsString(pattern), true);
+        LOG.info("Updated pattern output: {}", updatedPatternOutput);
+        // Normalize stop times.
+        JdbcTableWriter updateTripWriter = createTestTableWriter(tripsTable);
+        updateTripWriter.normalizeStopTimesForPattern(pattern.id, 0);
+        // Read pattern stops from database and check that the arrivals/departures have been updated.
+        JDBCTableReader<StopTime> stopTimesTable = new JDBCTableReader(Table.STOP_TIMES, testDataSource, testNamespace + ".", EntityPopulator.STOP_TIME);
```
evansiroky: Please format this line so it isn't so long.

src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
```diff
@@ -252,6 +254,34 @@ public String update(Integer id, String json, boolean autoCommit) throws SQLExce
...
     }
 
+    /**
+     * For a given pattern id and starting stop sequence (inclusive), normalize all stop times to match the pattern
+     * stops' travel times.
+     * @return number of stop times updated
+     */
+    public int normalizeStopTimesForPattern (int id, int beginWithSequence) throws SQLException {
+        try {
+            JDBCTableReader<PatternStop> patternStops = new JDBCTableReader(Table.PATTERN_STOP, dataSource, tablePrefix + ".", EntityPopulator.PATTERN_STOP);
```
evansiroky: Please format this line so it isn't so long.

src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
```diff
@@ -670,6 +702,73 @@ private int updateStopTimesForPatternStop(ObjectNode patternStop, int previousTr
...
+                "and t.pattern_id = ? " +
+                "and t.trip_id = st.trip_id",
+            timeField,
+            tablePrefix,
+            tablePrefix
+        );
+        if (serviceIdFilters != null && serviceIdFilters.size() > 0) {
+            getFirstTravelTimeSql += String.format(
+                " and t.service_id in (%s)",
+                Collections.nCopies(serviceIdFilters.size(), "?")
```
evansiroky: Does this work? Wouldn't this just create a bunch of question marks without commas? It doesn't appear that this filter is ever used anyways. Is it needed?

evansiroky: Also, there aren't any prepared statement set statements for these extra parameters either.

src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
```diff
@@ -670,6 +702,73 @@ private int updateStopTimesForPatternStop(ObjectNode patternStop, int previousTr
...
+        }
+        // Update stop times for individual trips with normalized travel times.
+        String updateTravelTimeSql = String.format(
+            "update %s.stop_times set arrival_time = ?, departure_time = ? where trip_id = ? and stop_sequence = ?",
+            tablePrefix
+        );
+        PreparedStatement updateStopTimeStatement = connection.prepareStatement(updateTravelTimeSql);
+        LOG.info(updateStopTimeStatement.toString());
+        final BatchTracker stopTimesTracker = new BatchTracker("stop_times", updateStopTimeStatement);
+        for (String tripId : timesForTripIds.keySet()) {
```
evansiroky: Could this for loop be avoided by doing the update statement using a join to find all applicable trip IDs?

evansiroky: Oh, hmm, I see why there are 2 loops. It could still probably be combined into one huge messy SQL query, but I suppose this is fine for now.

landonreed: 🤔... is `one huge messy SQL query` a desirable thing? I don't think it necessarily is.

evansiroky: It's only desirable if there are considerable performance benefits. However, that would entail much more investigation in running jmeter tests and whatnot so yeah, I'm fine with keeping this as is until users ever start complaining about the performance of this.

abyrd: And even if there are considerable performance benefits, it's only desirable if the code is factored to remain readable, and that performance gain is actually needed and a significant fraction of total runtime. Otherwise in my opinion the best thing to do is add a comment informing future readers of the alternative approach.

landonreed:
@evansiroky, I removed the service ID filter, which was really just a placeholder for a feature that I don't think we will implement. I also added another test that's not exactly related, but something that was missing and I wanted to verify.

Issue #233, opened 2019-05-21
Refactor PersistenceExpectations to contain multiple records


landonreed:
Currently `PersistenceExpectations` allows tests to verify that an imported GTFS table has certain values, but it can only do this for one value. It currently selects all records in a table and just checks that each row meets the various record expectations, which just doesn't seem right. This should be refactored so that multiple records can be provided for a given table and the db selection checks each row accordingly.

PR #236, opened 2019-05-24
Google Feed Validator (Near) Parity

landonreed:
### Checklist

### Description

Brings the validator library into (near) parity with the Google Feed Validator (re #167). The only thing remaining is better shape validation and some encoding/line ending stuff. This PR:
- improves the time zone error message
- checks that fare_attribute transfer_duration is zero if transfers are not permitted
- adds some stop time checks (ensures a stop time has a purpose and that timepoints have times)
- adds frequency overlap check
- adds supporting tests

This also refactors how ErrorExpectations are handled so it partially addresses #233.


src/main/java/com/conveyal/gtfs/validator/FrequencyValidator.java
```diff
@@ -0,0 +1,66 @@
...
+    private ListMultimap<String, Frequency> frequenciesById = ArrayListMultimap.create();
+
+    @Override
+    public void validate() {
+        // First, collect all frequencies for each trip ID.
+        for (Frequency frequency: feed.frequencies) frequenciesById.put(frequency.trip_id, frequency);
+        // Next iterate over each set of trip-specific frequency periods.
+        for (String tripId : frequenciesById.keySet()) {
+            List<Frequency> frequencies = frequenciesById.get(tripId);
+            if (frequencies.size() < 1) {
```
evansiroky: should this be <= 1?

src/main/java/com/conveyal/gtfs/validator/FrequencyValidator.java
```diff
@@ -0,0 +1,66 @@
...
+                // If there are not more than one frequencies defined for the trip, there can be no risk of overlapping
+                // frequency intervals.
+                return;
+            }
+            // Iterate over each frequency and check its period against the others for overlap.
+            for (int i = 0; i < frequencies.size(); i++) {
+                Frequency frequency = frequencies.get(i);
+                // Iterate over the other frequencies.
+                for (int j = 0; j < frequencies.size(); j++) {
```
evansiroky: This is needlessly comparing frequencies that were already compared. For example, this will compare Frequency 0 with Frequency 1 and then again will compare Frequency 1 with Frequency 0 which is the same thing. You can rewrite the first loop to terminate at `frequencies.size() - 1` and rewrite the 2nd loop as `for (int j = i + 1; j < frequencies.size(); j++) {`. However, you'll probably also want to add another check for "current wraps other".

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -456,10 +481,43 @@ private void assertThatImportedGtfsMeetsExpectations(
...
+                String entityId = rs.getString("entity_id");
+                String badValue = rs.getString("bad_value");
+                // Skip error expectation if not exists. But continue iteration to count all errors.
+                if (!errorExpectationIterator.hasNext()) continue;
+                ErrorExpectation errorExpectation = errorExpectationIterator.next();
+                LOG.info("Checking error {} (found={} expected={})", errorCount, errorType, errorExpectation.errorType);
+                assertThat(errorExpectation.errorType.toString(), equalTo(errorType));
+                if (errorExpectation.entityType != null) assertThat(errorExpectation.entityType, equalTo(entityType));
+                if (errorExpectation.entityId != null) assertThat(errorExpectation.entityId, equalTo(entityId));
+                // FIXME: How should we handle the expectation that the bad value expected matches null?
```
evansiroky: Use `org.hamcrest.core.IsNull.nullValue`. See https://stackoverflow.com/a/18987738/269834

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -456,10 +481,43 @@ private void assertThatImportedGtfsMeetsExpectations(
...
+                if (errorExpectation.badValue != null) assertThat(errorExpectation.badValue, equalTo(badValue));
+            }
+            if (errorExpectations.length < errorCount) {
+                LOG.warn("More errors stored in database ({}) than expectations ({}) found in test.", errorCount, errorExpectations.length);
```
evansiroky: Can you add an assertion for this condition? This should result in an error to make sure there are no false positives.

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -456,10 +481,43 @@ private void assertThatImportedGtfsMeetsExpectations(
...
                     foundRecord = true;
                     break;
+                } else {
+                    LOG.error("Persistence mismatch on record {}", numRecordsSearched);
                 }
             }
             assertThatPersistenceExpectationRecordWasFound(numRecordsSearched, foundRecord, fieldsWithMismatches);
         }
+        // Check that error expectations match errors stored in database.
+        if (errorExpectations != null && errorExpectations.length > 0) {
```
evansiroky: I think it'd be a good idea to also make sure there were zero errors reported if errorExpectations is null. That way we can be sure that clean feeds don't have errors.

src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -119,28 +128,39 @@ public void canLoadFeedWithBadDates () {
...
-                new ErrorExpectation[]{
-                    new ErrorExpectation("error_type", NewGTFSErrorType.TRIP_OVERLAP_IN_BLOCK.toString())
-                }
-            )
+    public void canLoadFeedWithErrors () {
+        PersistenceExpectation[] expectations = PersistenceExpectation.list();
+        ErrorExpectation[] errorExpectations = ErrorExpectation.list(
+            new ErrorExpectation(NewGTFSErrorType.FARE_TRANSFER_MISMATCH, "fare-02"),
+            new ErrorExpectation(NewGTFSErrorType.FREQUENCY_PERIOD_OVERLAP, "freq-01_09:00:00_to_10:00:00_every_10m00s"),
+            new ErrorExpectation(NewGTFSErrorType.FREQUENCY_PERIOD_OVERLAP, "freq-01_09:30:00_to_10:30:00_every_15m00s"),
```
evansiroky: I think it'd be good to add 2 more error expectations here to cover all possible overlaps:

Key: 

`|`: start or end of frequency
`-`: active time of frequency
`_`: frequency not active

A covers B:

A: `|---------|`
B: `___|--|____`

B covers A

A: `___|--|____`
B: `|---------|`

A starts before B starts, but A ends after B starts and A ends before B ends:

A: `_|----|____`
B: `____|-----|`

A starts after B starts, but A also starts before B ends

A: `____|-----|`
B: `_|----|____`

src/test/resources/fake-agency-overlapping-trips/frequencies.txt
```diff
@@ -0,0 +1,3 @@
+trip_id,start_time,end_time,headway_secs,exact_times
+freq-01,09:00:00,10:00:00,600,0
+freq-01,09:30:00,10:30:00,900,0
```
evansiroky: Can you add some more frequencies to this feed? Specifically, I think it'd make for a great test if we had all of the conditions recommended in https://github.com/conveyal/gtfs-lib/pull/236/files#r287460184. Also, it'd be great to add some valid frequency entries to make sure there are no false positives. ie `Freq A` is entirely before `Freq B` and `Freq A` is entirely after `Freq B`. Bonus points for adding even more edge cases like Freq A ends exactly when Freq B starts. Super bonus points for adding frequency entries that have interactions with daylight savings time transitions.

landonreed:
Thanks, @evansiroky. Just addressed the comments.

src/main/java/com/conveyal/gtfs/validator/FrequencyValidator.java
```diff
@@ -36,28 +36,38 @@ public void validate() {
...
-                        // Other frequency wraps current frequency.
-                        frequency.end_time <= prevFrequency.end_time && frequency.start_time >= prevFrequency.start_time ||
-                            // Frequency starts during other frequency.
-                            frequency.start_time >= prevFrequency.start_time && frequency.start_time < prevFrequency.end_time ||
-                            // Frequency ends during other frequency.
-                            frequency.end_time > prevFrequency.start_time && frequency.end_time <= prevFrequency.end_time
+                        // -- diagrams courtesy of esiroky --
+                        // A wraps B.
+                        // A: |---------|
+                        // B: ___|--|____
```
evansiroky: haha, nice. :)

PR #256, opened 2019-09-27
Change transfers#transfer_type column to ShortField

### LINKED ISSUE #255: Transfers#transfer_type should probably be a ShortField
## Observed behavior

The `transfer_type` column in the `transfers` table is defined as a `StringField`.

https://github.com/conveyal/gtfs-lib/blob/90d054cc610726203d4eaf94388ae0bc6b52387a/src/main/java/com/conveyal/gtfs/loader/Table.java#L255-L263

## Expected behavior

The `transfer_type` column should probably be a `ShortField`.

## Steps to reproduce the problem

n/a

## Notes on configuration and files required to reproduce the issue

n/a

evansiroky:
### Checklist

### Description

Fixes #255

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/Table.java
+++ b/src/main/java/com/conveyal/gtfs/loader/Table.java
@@ -256,7 +256,7 @@ public class Table {
             // FIXME: Do we need an index on from_ and to_stop_id
             new StringField("from_stop_id", REQUIRED).isReferenceTo(STOPS),
             new StringField("to_stop_id", REQUIRED).isReferenceTo(STOPS),
-            new StringField("transfer_type", REQUIRED),
+            new ShortField("transfer_type", REQUIRED, 3),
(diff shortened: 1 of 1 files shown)


evansiroky:
> LGTM. One thing though -- I can't recall, does changing the field type require any kind of database migration or issues with getting the field into the GraphQL schema?

There is no transfer type in the gtfs-lib GraphQL schema. 😨

landonreed:
@evansiroky, could you create an issue to add transfers.txt to the GraphQL schema (or add it as part of this PR) and verify whether or not a db migration needs to be run on databases to fix the field type?

evansiroky:
Here is the update SQL that would need to be ran: `ALTER TABLE %s.transfers ALTER COLUMN transfer_type TYPE INT USING transfer_type::integer`. It resulted in a bunch of errors because the transfers table isn't present in each namespace.

PR #269, opened 2020-01-22
Allow fetching of the priority of a validation error type

evansiroky:
### Checklist

### Description

This PR will expose the priority level of validation errors through the GraphQL schema. This is needed in order to implement https://github.com/ibi-group/datatools-ui/issues/477.


src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
```diff
@@ -421,6 +422,7 @@
             .field(MapFetcher.field("entity_id"))
             .field(MapFetcher.field("entity_sequence", GraphQLInt))
             .field(MapFetcher.field("bad_value"))
+            .field(ErrorPriorityFetcher.field("error_type_priority"))
```
landonreed: It seems a little bit overkill to define a custom Fetcher for the `priority` string. It looks like every other string field is being fetched with the `MapFetcher`. Can you describe your thinking here? It seems like unneeded complexity and like we're just introducing a new pattern.

landonreed: I think the intent with the `errorCountType` definition below is to contain high-level information about the error type, including the count, english message and priority fields. I think this fetcher should be removed because its redundant information. Open to hearing a counter argument though.

evansiroky: I originally thought each error needed this info, but after implementing https://github.com/ibi-group/datatools-ui/pull/527 I realized I didn't really need it. So yeah, I'll remove it.

Issue #130, opened 2018-08-01
Better shape validation


landonreed:
Recently some shapes created in the editor were discarded during graph building by OTP.  I think we should add some functionality similar to the [`GTFSPatternHopFactory#isValid`](https://github.com/opentripplanner/OpenTripPlanner/blob/452d7739fd8c3ab0ae93579cb10fd22bea79b1f7/src/main/java/org/opentripplanner/routing/edgetype/factory/GTFSPatternHopFactory.java#L1177-L1201) method.

abyrd:
Should the editor even be capable of creating invalid shapes? What was wrong with the shapes it produced which caused them to be rejected by OTP?

I agree we want to sanity check inputs an outputs as much as possible. My main concern would be maintaining duplicate code - is it possible to design these checks such that they can be reused in GTFS-lib and other projects?

PR #276, opened 2020-04-02
MTC Field length validation

binh-dam-ibigroup:
### Checklist

- [na] The description lists all applicable issues this PR seeks to resolve
- [na] The description lists any configuration setting(s) that differ from the default settings

### Description

This PR adds the following (in pursuit of ibi-group/datatools-ui#544):
* Ability to specify custom validator classes for method `GTFS.validate()` and related methods.
* A custom validator `MTCValidator` for custom validation of 511 MTC guidelines on GTFS field length.
* An entry in `NewGTFSErrorType` for field values that are too long. 
* Unit tests and a database loading/validation test case that uses MTCValidator.


evansiroky:
Unassigning myself until #277 gets merged.

src/main/java/com/conveyal/gtfs/loader/Feed.java
```diff
@@ -65,44 +58,51 @@ public Feed (DataSource dataSource, String tablePrefix) {
         routes = new JDBCTableReader(Table.ROUTES, dataSource, tablePrefix, EntityPopulator.ROUTE);
         stops = new JDBCTableReader(Table.STOPS, dataSource, tablePrefix, EntityPopulator.STOP);
         trips = new JDBCTableReader(Table.TRIPS, dataSource, tablePrefix, EntityPopulator.TRIP);
-//        shapePoints = new JDBCTableReader(Table.SHAPES, dataSource, tablePrefix, EntityPopulator.SHAPE_POINT);
```
evansiroky, line 68: I realize this has been commented out for a while since @landonreed committed https://github.com/conveyal/gtfs-lib/commit/9244ff072cc6aa85bf6d6d315afa7861efdb10d4. But why? Should we add it back?

src/main/java/com/conveyal/gtfs/validator/MTCValidator.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Checks that the length of a string (or Object.toString()) does not exceed a length.
+     * Reports an error if the length is exceeded.
+     * @param entity The containing GTFS entity (for error reporting purposes).
+     * @param objValue The value to check.
+     * @param maxLength The length to check, should be positive or zero.
+     * @return true if the length of objValue.toString() is maxLength or less or if objValue is null; false otherwise.
+     */
+    public boolean fieldLengthShouldNotExceed(Entity entity, Object objValue, int maxLength) {
```
evansiroky, line 48: Can we rename this `validFieldLength`?

src/main/java/com/conveyal/gtfs/validator/ReversedTripValidator.java
```diff
@@ -50,7 +50,8 @@ public boolean validate(Feed feed, boolean repair) {
             if (trip.shape_id == null) {
                 isValid = false;
                 if (missingShapeErrorCount < errorLimit) {
-                    feed.errors.add(new MissingShapeError(trip));
+                    // FIXME store MissingShape errors
+                    // feed.errors.add(new MissingShapeError(trip));
```
evansiroky, line 54: What's with all these changes related to shape errors?

landonreed: When working on this, I discovered that the `Feed#errors` property was being used by some other validators (including `ReversedTripValidator`) to store errors that were never making it into the database. We have an open issue (#130) that somewhat covers this already, although I'm not sure if we have a good plan of attack for how to restore these checks.

src/main/java/com/conveyal/gtfs/loader/Feed.java
landonreed: Perhaps we should. It might be worth revisiting #130.

src/main/java/com/conveyal/gtfs/error/NewGTFSErrorType.java
```diff
@@ -11,6 +11,7 @@
     URL_FORMAT(Priority.MEDIUM, "URL format should be <scheme>://<authority><path>?<query>#<fragment>"),
     LANGUAGE_FORMAT(Priority.LOW, "Language should be specified with a valid BCP47 tag."),
     ILLEGAL_FIELD_VALUE(Priority.MEDIUM, "Fields may not contain tabs, carriage returns or new lines."),
+    FIELD_VALUE_TOO_LONG(Priority.MEDIUM, "Field value has too many characters."),
```
landonreed, line 14: @binh-dam-ibigroup, can you add a comment above this new type describing that this is an MTC-specific issue type (for now). Another approach would be to add another Priority enum (e.g. `Priority.MTC`), but I'm not 100% on that approach.

cc @evansiroky

src/main/java/com/conveyal/gtfs/validator/MTCValidator.java
```diff
@@ -0,0 +1,56 @@
...
+import com.conveyal.gtfs.error.SQLErrorStorage;
+import com.conveyal.gtfs.loader.Feed;
+import com.conveyal.gtfs.model.*;
+
+import static com.conveyal.gtfs.error.NewGTFSErrorType.FIELD_VALUE_TOO_LONG;
+
+/**
+ * MTCValidator checks in a GTFS feed that the length of certain field values
+ * do not exceed the 511 MTC guidelines. (TODO: add guidelines URL.)
+ * To refer to specific limits, search the guidelines for the word 'character'.
```
landonreed, line 12: Can you adjust this to say something like "MTCValidator runs a set of custom validation checks for GTFS feeds managed by MTC in Data Tools. " and then list the field length check as the primary check for now? Also, might be worth referring to the issue: https://github.com/ibi-group/datatools-ui/issues/544

src/main/java/com/conveyal/gtfs/error/NewGTFSErrorType.java
binh-dam-ibigroup: I'll add the comment for MTC. I also think error types should remain generic if another agency wants to use them.

src/main/java/com/conveyal/gtfs/validator/ReversedTripValidator.java
binh-dam-ibigroup: Is that an MTC-related need?

src/main/java/com/conveyal/gtfs/validator/MTCValidator.java
binh-dam-ibigroup: Perhaps... however the max length requirement will be less obvious from the MTCValidator class. What is @landonreed's opinion?

landonreed: I'm with @evansiroky on this one.

landonreed: Also, I don't love the idea of accepting a generic `Object` as an arg here. Can we just have two methods one for String and one for URL (URL can perhaps call the String method)?

src/main/java/com/conveyal/gtfs/validator/MTCValidator.java
```diff
@@ -0,0 +1,56 @@
...
+        for (Stop stop : feed.stops) {
+            fieldLengthShouldNotExceed(stop, stop.stop_name, 100);
+        }
+
+        for (Trip trip : feed.trips) {
+            fieldLengthShouldNotExceed(trip, trip.trip_headsign, 120);
+            fieldLengthShouldNotExceed(trip, trip.trip_short_name, 50);
+        }
+
+        // TODO: Handle calendar_attributes.txt?
```
landonreed, line 37: I don't think this note needs to be here does it?

src/main/java/com/conveyal/gtfs/validator/MTCValidator.java
binh-dam-ibigroup: Good point.

src/main/java/com/conveyal/gtfs/validator/ReversedTripValidator.java
binh-dam-ibigroup: Good one.

binh-dam-ibigroup:
PR requests addressed, and the tests are passing now... Please review.

src/main/java/com/conveyal/gtfs/error/NewGTFSErrorType.java
```diff
@@ -11,6 +11,7 @@
     URL_FORMAT(Priority.MEDIUM, "URL format should be <scheme>://<authority><path>?<query>#<fragment>"),
     LANGUAGE_FORMAT(Priority.LOW, "Language should be specified with a valid BCP47 tag."),
     ILLEGAL_FIELD_VALUE(Priority.MEDIUM, "Fields may not contain tabs, carriage returns or new lines."),
+    // The error type below is an MTC-specific requirement.
```
evansiroky, line 14: Let's organize these into sections instead of having an individual comment for potentially each error type.

binh-dam-ibigroup: Did you have anything in mind for reorganizing?

evansiroky: I'd say one section for errors that apply during all types of validation and another section for MTC-specific errors.

binh-dam-ibigroup: How about this? https://github.com/conveyal/gtfs-lib/pull/276/files#diff-8946752ed48c4dd733a37388eec80ac8

PR #291, opened 2020-09-30
Silence number excessive parse exceptions logs

landonreed:
### Checklist

### Description

There are a large number of useless exceptions being output to the log when adding NewGtfsError entries that do not have sequence values. This attempts to silence most of those. re ibi-group/datatools-server#342

### DIFF
--- a/src/main/java/com/conveyal/gtfs/error/NewGTFSError.java
+++ b/src/main/java/com/conveyal/gtfs/error/NewGTFSError.java
@@ -2,6 +2,8 @@ package com.conveyal.gtfs.error;
 
 import com.conveyal.gtfs.loader.Table;
 import com.conveyal.gtfs.model.Entity;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


src/main/java/com/conveyal/gtfs/error/NewGTFSError.java
```diff
@@ -111,15 +113,23 @@ public NewGTFSError setEntityId(String entityId) {
...
+     * Builder to add entity sequence info from string (values during load stage are passed in as strings from csv
+     * reader).
+     */
     public NewGTFSError setSequence(String sequenceAsString) {
-        try {
-            // Parse int from string value found during load stage.
-            this.entitySequenceNumber = Integer.parseInt(sequenceAsString);
-        } catch (NumberFormatException e) {
-            e.printStackTrace();
+        if (sequenceAsString == null || sequenceAsString.isEmpty()) {
```
evansiroky, line 121: I looked at the internal workings of `String.isEmpty` and also `String.isBlank` and I think `String.isBlank` would be a better choice here since `isEmpty` only checks if the String is a zero-length string whereas isBlank will make sure that both the string isn't zero-length, but also contains at least one non-whitespace character.

evansiroky: On second thought, `isEmpty` actually seems to comply with the [GTFS spec](http://gtfs.org/reference/static/#file-requirements):

> Remove any extra spaces between fields or field names. Many parsers consider the spaces to be part of the value, which may cause errors.

evansiroky (approved):
I'm going to fast-track this to a release.

PR #297, opened 2020-11-30
Add graphql fetcher for shapes as encoded polylines

landonreed:
re ibi-group/datatools-ui#627

### Checklist

### Description

Adds encoded polyline utils (from Mapbox repo) and GraphQL query for shapes as encoded polylines.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
+++ b/src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
@@ -166,7 +166,7 @@ public class GraphQLGtfsSchema {
             .build();
 
     // Represents a set of rows from shapes.txt joined by shape_id
-    public static final GraphQLObjectType shapeType = newObject().name("shape")
+    public static final GraphQLObjectType shapeEncodedPolylineType = newObject().name("shapeEncodedPolyline")
         .field(string("shape_id"))
         .field(string("polyline"))
         .build();
@@ -652,16 +652,8 @@ public class GraphQLGtfsSchema {
                     .dataFetcher(new JDBCFetcher("patterns"))
                     .build())
             .field(newFieldDefinition()
-                .name("shapes")
-                .type(new GraphQLList(shapeType))
-                .argument(intArg(ID_ARG))
-                .argument(intArg(LIMIT_ARG))
-                .argument(intArg(OFFSET_ARG))
-                .argument(floatArg(MIN_LAT))
-                .argument(floatArg(MIN_LON))
-                .argument(floatArg(MAX_LAT))
-                .argument(floatArg(MAX_LON))
-                .argument(multiStringArg("shape_id"))
+                .name("shapes_as_polylines")
+                .type(new GraphQLList(shapeEncodedPolylineType))
                 // DataFetchers can either be class instances implementing the interface, or a static function reference
                 .dataFetcher(new PolylineFetcher())
                 .build())
--- a/src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
+++ b/src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
@@ -5,6 +5,7 @@ import com.conveyal.gtfs.graphql.fetchers.FeedFetcher;
 import com.conveyal.gtfs.graphql.fetchers.JDBCFetcher;
 import com.conveyal.gtfs.graphql.fetchers.MapFetcher;
 import com.conveyal.gtfs.graphql.fetchers.NestedJDBCFetcher;
+import com.conveyal.gtfs.graphql.fetchers.PolylineFetcher;
 import com.conveyal.gtfs.graphql.fetchers.RowCountFetcher;
 import com.conveyal.gtfs.graphql.fetchers.SQLColumnFetcher;
 import com.conveyal.gtfs.graphql.fetchers.SourceObjectFetcher;
@@ -163,6 +164,12 @@ public class GraphQLGtfsSchema {
             .field(MapFetcher.field("shape_pt_sequence", GraphQLInt))
             .field(MapFetcher.field("point_type", GraphQLInt))
             .build();
+
+    // Represents a set of rows from shapes.txt joined by shape_id
+    public static final GraphQLObjectType shapeType = newObject().name("shape")
+        .field(string("shape_id"))
+        .field(string("polyline"))
+        .build();
 
 
     // Represents rows from frequencies.txt
@@ -644,6 +651,20 @@ public class GraphQLGtfsSchema {
                     // DataFetchers can either be class instances implementing the interface, or a static function reference
                     .dataFetcher(new JDBCFetcher("patterns"))
                     .build())
+            .field(newFieldDefinition()
+                .name("shapes")
+                .type(new GraphQLList(shapeType))
+                .argument(intArg(ID_ARG))
+                .argument(intArg(LIMIT_ARG))
+                .argument(intArg(OFFSET_ARG))
+                .argument(floatArg(MIN_LAT))
+                .argument(floatArg(MIN_LON))
+                .argument(floatArg(MAX_LAT))
+                .argument(floatArg(MAX_LON))
+                .argument(multiStringArg("shape_id"))
+                // DataFetchers can either be class instances implementing the interface, or a static function reference
+                .dataFetcher(new PolylineFetcher())
+                .build())
             // Then the fields for the sub-tables within the feed (loaded directly from GTFS).
             .field(newFieldDefinition()
                     .name("agency")
@@ -798,8 +819,6 @@ public class GraphQLGtfsSchema {
             .newSchema()
             .query(feedQuery)
 //            .query(patternsForStopQuery)
-            // TODO: Add mutations.
-            // .mutation(someMutation)
             .build();
 
 
--- a/src/main/java/com/conveyal/gtfs/graphql/fetchers/PolylineFetcher.java
+++ b/src/main/java/com/conveyal/gtfs/graphql/fetchers/PolylineFetcher.java
@@ -12,12 +12,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.sql.Connection;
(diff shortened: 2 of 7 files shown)


src/main/java/com/conveyal/gtfs/util/PolylineUtils.java
```diff
@@ -0,0 +1,325 @@
+package com.conveyal.gtfs.util;
+
+import com.sun.istack.internal.NotNull;
```
evansiroky, line 3: This is causing a compilation error in the [Travis build](https://travis-ci.com/github/conveyal/gtfs-lib/builds/205469968).

src/test/java/com/conveyal/gtfs/util/PolylineUtilsTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+
+import static com.conveyal.gtfs.util.PolylineUtils.decode;
+import static com.conveyal.gtfs.util.PolylineUtils.encode;
+import static com.conveyal.gtfs.util.PolylineUtils.simplify;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class PolylineUtilsTest {
```
evansiroky, line 19: Is this test case also taken from mapbox-java? If so we should properly attribute.

src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
```diff
@@ -644,6 +651,20 @@
                     // DataFetchers can either be class instances implementing the interface, or a static function reference
                     .dataFetcher(new JDBCFetcher("patterns"))
                     .build())
+            .field(newFieldDefinition()
+                .name("shapes")
```
evansiroky, line 655: To avoid confusion with existing schema items with the `shape(s)` name (such as `tripType#shape`), perhaps it would be good to rename this `shape_encoded_polyline` or something.

src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
```diff
@@ -644,6 +651,20 @@
...
+                .type(new GraphQLList(shapeType))
+                .argument(intArg(ID_ARG))
+                .argument(intArg(LIMIT_ARG))
+                .argument(intArg(OFFSET_ARG))
+                .argument(floatArg(MIN_LAT))
+                .argument(floatArg(MIN_LON))
+                .argument(floatArg(MAX_LAT))
+                .argument(floatArg(MAX_LON))
+                .argument(multiStringArg("shape_id"))
```
evansiroky, line 664: Do any of these get used in the PolylineFetcher? If not, they should be removed.

src/main/java/com/conveyal/gtfs/graphql/fetchers/PolylineFetcher.java
```diff
@@ -0,0 +1,87 @@
...
+        Connection connection = null;
+        try {
+            connection = GTFSGraphQL.getConnection();
+            Statement statement = connection.createStatement();
+            String sql = String.format(
+                "select shape_id, shape_pt_lon, shape_pt_lat from %s.shapes order by shape_id",
+                namespace
+            );
+            LOG.info("SQL: {}", sql);
+            if (statement.execute(sql)) {
```
evansiroky, line 43: Seems like this will always fetch every single shape and not do any filtering according to shape_id or bounding box. Either that logic should be implemented here, or the args in the GraphQL schema should be removed.

evansiroky (changes requested):
- Fix Travis Build
- Consider renaming new `shapes` schema name to something that won't confuse it with the raw GTFS shapes table.
- Resolve discrepancy between GraphQL schema that implies advanced filtering of shapes versus actual code seeming to provide every possible shape.

src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
```diff
@@ -644,6 +651,12 @@
                     // DataFetchers can either be class instances implementing the interface, or a static function reference
                     .dataFetcher(new JDBCFetcher("patterns"))
                     .build())
+            .field(newFieldDefinition()
+                .name("shapes_as_polylines")
```
evansiroky, line 655: Nit: would like this named `shapes_as_encoded_polylines`.

src/main/java/com/conveyal/gtfs/graphql/fetchers/PolylineFetcher.java
```diff
@@ -0,0 +1,87 @@
...
+        GeometryFactory gf = new GeometryFactory();
+        List<Shape> shapes = new ArrayList<>();
+        Map<String, Object> parentFeedMap = environment.getSource();
+        String namespace = (String) parentFeedMap.get("namespace");
+        Connection connection = null;
+        try {
+            connection = GTFSGraphQL.getConnection();
+            Statement statement = connection.createStatement();
+            String sql = String.format(
+                "select shape_id, shape_pt_lon, shape_pt_lat from %s.shapes order by shape_id",
```
evansiroky, line 39: Shouldn't this also be ordered by `shape_pt_sequence`?

Also, I wonder if the above query would work better if it were split into 2 queries to avoid excessive sorting. The first query could get the distinct shape_id's and then the next query would query all shape lat/lons for a specific shape_id.

evansiroky (changes requested):
See two non-blocking comments.

However, has this been tested with a massive feed that has caused datatools-server to crash in the past? This should be done to make sure this can run without encountering an out-of-memory problem or some other issue and also to make sure the performance is acceptable. Once that is done, please share the results.

If the performance is not acceptable, perhaps there is a way to pre-compute the encoded polylines, although that could get tricky to maintain in editor databases where the shapes can be updated.

src/main/java/com/conveyal/gtfs/graphql/GraphQLGtfsSchema.java
landonreed: I think the format of the polylines is fairly obvious by looking at the response?

src/main/java/com/conveyal/gtfs/graphql/fetchers/PolylineFetcher.java
landonreed: Yea, probably should be ordered by `shape_pt_sequence` as well. Good catch. I don't think breaking up the query will do anything here. There are indexes on both fields (shape_id and sequence).

src/main/java/com/conveyal/gtfs/graphql/fetchers/PolylineFetcher.java
```diff
@@ -21,6 +21,15 @@
...
  * GraphQL fetcher to get encoded polylines for all shapes in a GTFS feed.
+ *
+ * Note: this fetcher was developed to prevent out of memory errors associated with requesting all trip patterns with
+ * their associated shapes. Previously, attempting to fetch all shapes for a large feed would require a separate SQL
+ * query for each shape (to join on shape_id) and result in many duplicated shapes in the response.
+ *
+ * Running a shapes polyline query on a 2017 Macbook Pro (2.5 GHz Dual-Core Intel Core i7) results in the following:
+ * - NL feed (11291 shapes): ~12 seconds (but nobody should be editing such a large feed in the editor)
+ * - TriMet (1302 shapes): ~8 seconds
+ * - MBTA (1272 shapes): ~4 seconds
```
evansiroky, line 32: Good to know, but a response time of 4 seconds is definitely something that could be improved upon.

src/main/java/com/conveyal/gtfs/graphql/fetchers/PolylineFetcher.java
```diff
@@ -36,7 +45,7 @@ public Object get(DataFetchingEnvironment environment) {
             connection = GTFSGraphQL.getConnection();
             Statement statement = connection.createStatement();
             String sql = String.format(
-                "select shape_id, shape_pt_lon, shape_pt_lat from %s.shapes order by shape_id",
+                "select shape_id, shape_pt_lon, shape_pt_lat from %s.shapes order by shape_id, shape_pt_sequence",
```
evansiroky, line 48: If 4 seconds is tolerable for now, then this is fine to keep, but this single query across all records could be some good low hanging fruit to refactor since it likely performs an excessive amount of sorting. 

Let's say there are 100 shapes. Let's say each shape has 500 points. 

This current implementation would sort all 100 shapes and the 500 shape points within each, so 50,000 items to sort once. Sorting takes n log<sub>2</sub>(n) time, so 1 * 50000 * log<sub>2</sub>(50000) = 780k operations. 

If there were 100 separate queries, but each only sorted 500 items at a time, this is substantially less calculation. 100 * 500 * log<sub>2</sub>(500) = 448k operations.

evansiroky (approved):
Good enough for now, but I still recommend looking into breaking up the shapes query into 2 separate queries.

src/main/java/com/conveyal/gtfs/graphql/fetchers/PolylineFetcher.java
evansiroky: There are actually 719k lines in the TriMet shapes.txt file with 1,909 unique shape IDs (so about 377 points per shape on average). With that, the numbers become:

- 1 * 719k * log<sub>2</sub>(719k) = 13.9m operations
- 1909 * 377 * log<sub>2</sub>(377) = 6.1m operations

landonreed: I just tried splitting up the queries into a `select distinct shape_id` and then iterating over the ids with a `select shape_id, shape_pt_lon, shape_pt_lat from %s.shapes where shape_id = ? order by shape_pt_sequence` and it appears to take longer for the NL feed (between 14-17 seconds). This is kind of what I was trying to get at previously: this is such a common SQL operation (a query that orders by two indexed fields) that it's probably already sufficiently optimized internally better than any kind of fanciness we could put on top of it.

PR #312, opened 2021-04-21
GTFS Spec Updates

br648:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This PR seeks to address the changes that have been made to the GTFS spec between Augusut 2018 and present. In particular, this PR is looking to enhance GTFS-lib by including conditionally required field checks which are now part of the GTFS spec. Initial work to determine the required updates can be reviewed here:

https://docs.google.com/document/d/1J_RlOx6wUml_fGhxX6pW1dr1-JKPMInia-73FTr3LJ8/edit

### Current Breaking Changes

Routes -> continuous_pickup (New field)
Routes -> continuous_drop_off (New field)
Stop Times -> continuous_pickup (New field)
Stop Times -> continuous_drop_off (New field)
Stop -> platform_code (New field)
Stop -> wheelchair_boarding (Type change from String to Int)
Feed Info -> default_lang (New field)
Feed Info -> feed_contact_email (New field)
Feed Info -> feed_contact_url (New field)

PR #313, opened 2021-04-21
refactor: re-order stops/fare_rules for ref check

landonreed:
Provides changes for #312.

@br648, something is broken with the parent_station test. I can't quite figure out what, but maybe your fresh eyes will spot it.


br648:
@landonreed regarding the parent_station check issue. All checks have to be carried out as the range is different for this field. I have updated.

PR #347, opened 2022-03-03
Flex validation

br648:
### Checklist

### Description

This PR covers all the spec flex validation items selected from the field descriptions on this page: https://github.com/MobilityData/gtfs-flex/blob/master/spec/reference.md


src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    public FlexValidator(Feed feed, SQLErrorStorage errorStorage) {
+        super(feed, errorStorage);
+    }
+
+    @Override
+    public void validate() {
+        feed.locationGroups.forEach(locationGroup ->
+            validateLocationGroups(locationGroup, Lists.newArrayList(feed.stops), Lists.newArrayList(feed.locations))
```
binh-dam-ibigroup: I am wondering about the need to pass fields from `feed` to the `validate*` methods.
I think those should be retrieved directly from within the method bodies.
- These methods are instance methods.
- These methods access the `errors` field.
- `feed` is a field.

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,358 @@
...
+        );
+        feed.stopTimes.forEach(stopTime ->
+            validateStopTimes(stopTime, Lists.newArrayList(feed.locationGroups), Lists.newArrayList(feed.locations))
+        );
+        feed.bookingRules.forEach(this::validateBookingRules);
+        // Register errors, if any, once all checks have been completed.
+        errors.forEach(this::registerError);
+    }
+
+    public void validateLocationGroups(LocationGroup locationGroup, List<Stop> stops, List<Location> locations) {
```
binh-dam-ibigroup: Rename to singular (only one location group gets validated).

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,358 @@
...
+            locations.stream().anyMatch(location -> location.location_id.equals(locationGroup.location_group_id))) {
+            // Location group id must not match a location id.
+            errors.add(NewGTFSError.forEntity(
+                locationGroup,
+                NewGTFSErrorType.FLEX_FORBIDDEN_LOCATION_GROUP_ID).setBadValue(locationGroup.location_group_id)
+            );
+        }
+    }
+
+    public void validateLocations(Location location, List<Stop> stops, List<FareRule> fareRules) {
```
binh-dam-ibigroup: Rename to singular (only one location is validated).

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
binh-dam-ibigroup: The other approach is to make each of these methods static, each returning a set of errors that get merged in `@Override void validate()`.

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,358 @@
...
+        ) {
+            // zone id is required if fare rules are defined.
+            errors.add(NewGTFSError.forEntity(
+                location,
+                NewGTFSErrorType.FLEX_MISSING_FARE_RULE).setBadValue(location.zone_id)
+            );
+        }
+    }
+
+    public void validateStopTimes(
```
binh-dam-ibigroup: Rename to singular.

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,358 @@
...
+            // safe_duration_offset is forbidden if stop_id does not refer to a location group or location.
+            errors.add(NewGTFSError.forEntity(
+                    stopTime,
+                    NewGTFSErrorType.FLEX_FORBIDDEN_SAFE_DURATION_OFFSET)
+                .setBadValue(Double.toString(stopTime.safe_duration_offset))
+            );
+        }
+    }
+
+    public void validateBookingRules(BookingRule bookingRule) {
```
binh-dam-ibigroup: Rename to singular.

src/test/java/com/conveyal/gtfs/validator/FlexValidatorTest.java
```diff
@@ -0,0 +1,563 @@
...
+            (LocationGroup) locationGroupArguments.testObject,
+            locationGroupArguments.stops,
+            locationGroupArguments.locations
+        );
+        checkValidationErrorsMatchExpectedErrors(flexValidator.errors, locationGroupArguments.expectedErrors);
+    }
+
+    private static Stream<Arguments> createLocationGroupChecks() {
+        return Stream.of(
+            Arguments.of(
```
binh-dam-ibigroup: You don't need to use `Arguments.of` if  only passing a single argument to each test.
(Apply to the other parametrized tests too)
```java
private static Stream<LocationGroupArguments> createLocationGroupChecks() {
    return Stream.of(
        new LocationGroupArguments(...),
        new LocationGroupArguments(...),
        ...
    );    
```

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,358 @@
...
+        feed.stopTimes.forEach(stopTime ->
+            validateStopTimes(stopTime, Lists.newArrayList(feed.locationGroups), Lists.newArrayList(feed.locations))
+        );
+        feed.bookingRules.forEach(this::validateBookingRules);
+        // Register errors, if any, once all checks have been completed.
+        errors.forEach(this::registerError);
+    }
+
+    public void validateLocationGroups(LocationGroup locationGroup, List<Stop> stops, List<Location> locations) {
+        if (!stops.isEmpty() && stops.stream().anyMatch(stop -> stop.stop_id.equals(locationGroup.location_group_id))) {
```
binh-dam-ibigroup: If not passing fields from `feed`, could something like `StreamSupport.stream(feed.stops.spliterator(), false)` work, or would it be too clunky?

binh-dam-ibigroup: This `if` condition is important logic for the validator, it should be extracted to a method `locationGroupIdIsStopId(String locationGroupId)` that utilizes `feed.stops` (or a more generic version of that) and lets you delete the comment below.

Same applies at least to the other checks in this class that use `anyMatch`.

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    public void validateStopTimes(
+        StopTime stopTime,
+        List<LocationGroup> locationGroups,
+        List<Location> locations) {
+
+        if (stopTime.arrival_time != INT_MISSING &&
+            (stopTime.start_pickup_dropoff_window != INT_MISSING ||
+                stopTime.end_pickup_dropoff_window != INT_MISSING)
```
binh-dam-ibigroup, line 149: Consider extracting a variable or method to replace the comments inside the if block.

binh-dam-ibigroup (changes requested):
First round of comments - mainly refactoring tests and validation checks.

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,358 @@
...
+        this(null, null);
+    }
+
+    public FlexValidator(Feed feed, SQLErrorStorage errorStorage) {
+        super(feed, errorStorage);
+    }
+
+    @Override
+    public void validate() {
+        feed.locationGroups.forEach(locationGroup ->
```
binh-dam-ibigroup: There needs to be a check on whether a feed contains flex stuff, or that the corresponding tables are present, before this is allowed to proceed. When loading a large set of feeds without flex data, some of this code is repeatedly trying to hit the locations table and as a result takes very long to complete (or goes in an infinite loop?) while consuming a lot of CPU power. Running this locally, I had to kill the datatools-server process after a few minutes.

binh-dam-ibigroup (changes requested):

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
br648: I went with the second option so the errors are still available for testing.

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
br648: Sorry about that. I forgot to add the check.

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
br648: This might bulk the code unnecessarily if I'm to do it for all similar `if` statements, so I'm in two minds. I think the comments are useful and some of the detail will be lost if converted to a variable. I guess if I went with methods I could put the comment in the method comment, but I'm then back to my original point. If you think this is a show stopper, I'm happy to make the changes, if not, I'm leaning towards leaving it.

src/main/java/com/conveyal/gtfs/error/NewGTFSErrorType.java
```diff
@@ -24,6 +24,34 @@
...
+    FLEX_FORBIDDEN_PRIOR_START_TIME(Priority.HIGH, "It is forbidden to define a prior notice start time if prior notice start day is not defined."),
+    FLEX_REQUIRED_END_PICKUP_DROPOFF_WINDOW(Priority.HIGH, "End pickup drop off window is required if a stop id refers to a location group or location."),
+    FLEX_REQUIRED_PRIOR_NOTICE_DURATION_MIN(Priority.HIGH, "Prior notice last day is required for booking type 2 (Up to prior day(s) booking)."),
+    FLEX_REQUIRED_PRIOR_NOTICE_LAST_DAY(Priority.HIGH, "Prior notice last day is required if a stop id refers to a location group or location."),
+    FLEX_REQUIRED_PRIOR_NOTICE_START_TIME(Priority.HIGH, "Prior notice start time is required if prior notice start day is defined."),
+    FLEX_REQUIRED_START_PICKUP_DROPOFF_WINDOW(Priority.HIGH, "Prior notice duration min is required for booking type 1 (Up to same-day booking with advance notice)."),
+    FLEX_SAVE_FACTORS_EXCEEDED(Priority.HIGH, "Safe duration factor + safe duration offset plus be greater than mean duration factor + mean duration offset."),
```
binh-dam-ibigroup: Typo ~Save~ => **Safe**.

binh-dam-ibigroup: Remove "plus" in "safe duration offset ~plus~ **should** be greater..."

src/main/java/com/conveyal/gtfs/error/NewGTFSErrorType.java
```diff
@@ -107,6 +108,7 @@
     TRIP_NEVER_ACTIVE(Priority.MEDIUM, "A trip is defined, but its service is never running on any date."),
     TRIP_OVERLAP_IN_BLOCK(Priority.MEDIUM, "A trip overlaps another trip and shares the same block_id."),
     TRIP_TOO_FEW_STOP_TIMES(Priority.MEDIUM, "A trip must have at least two stop times to represent travel."),
+    TRIP_SPEED_NOT_VALIDATED(Priority.MEDIUM, "Trip speed not validated because it contains at least one location or location group."),
```
binh-dam-ibigroup: Maybe this can be a low priority message?

src/main/java/com/conveyal/gtfs/validator/SpeedTripValidator.java
```diff
@@ -37,18 +38,26 @@
...
+        Route route,
+        List<StopTime> stopTimes,
+        List<Stop> stops,
+        List<Location> locations,
+        List<LocationGroup> locationGroups
+    ) {
+        if (hasLocationOrLocationGroupReferences(locations, locationGroups)) {
+            registerError(NewGTFSError.forFeed(
+                NewGTFSErrorType.TRIP_SPEED_NOT_VALIDATED,
+                "Trip speed not validated because it contains at least one location or location group.")
```
binh-dam-ibigroup: If validating each entity, the trip id and line number should be included in this error report. Otherwise in the UI you see a potentially long list that looks like



Line # | ID | Bad value
-- | -- | --
N/A | N/A | Trip speed not validated because it contains at least one location or location group.
N/A | N/A | Trip speed not validated because it contains at least one location or location group.
... | ... | ...

src/main/java/com/conveyal/gtfs/validator/SpeedTripValidator.java
```diff
@@ -37,18 +38,26 @@
...
+    public void validateTrip(
+        Trip trip,
+        Route route,
+        List<StopTime> stopTimes,
+        List<Stop> stops,
+        List<Location> locations,
+        List<LocationGroup> locationGroups
+    ) {
+        if (hasLocationOrLocationGroupReferences(locations, locationGroups)) {
+            registerError(NewGTFSError.forFeed(
```
binh-dam-ibigroup: As written, it should be `NewGTFSError.forEntity`. If it needs to be for the feed, this check needs to be at feed level (e.g. in `FlexValidator.validate()`).

binh-dam-ibigroup (changes requested):
I'd recommend moving the trip speed check to be at feed level as I think was the intention. Also fix typos.

src/main/java/com/conveyal/gtfs/validator/SpeedTripValidator.java
br648: Line number is not available, but have updated to include the trip entity and trip id.

src/main/java/com/conveyal/gtfs/error/NewGTFSErrorType.java
br648: Fixed in https://github.com/conveyal/gtfs-lib/pull/347/commits/da910db5c20ff9e8f6ef9f25611876e7abaf5501

src/main/java/com/conveyal/gtfs/error/NewGTFSErrorType.java
br648: Updated in https://github.com/conveyal/gtfs-lib/pull/347/commits/da910db5c20ff9e8f6ef9f25611876e7abaf5501

src/main/java/com/conveyal/gtfs/validator/SpeedTripValidator.java
br648: Addressed in https://github.com/conveyal/gtfs-lib/pull/347/commits/da910db5c20ff9e8f6ef9f25611876e7abaf5501

src/main/java/com/conveyal/gtfs/validator/SpeedTripValidator.java
br648: Fixed in https://github.com/conveyal/gtfs-lib/pull/347/commits/da910db5c20ff9e8f6ef9f25611876e7abaf5501

src/main/java/com/conveyal/gtfs/validator/NewTripTimesValidator.java
```diff
@@ -47,13 +47,16 @@
 
     // As an optimization, these validators are fed the stoptimes for each trip to avoid repeated iteration and grouping.
     private final TripValidator[] tripValidators;
+    private final TripValidator[] tripValidatorsWithAdditionalProcesses;
```
binh-dam-ibigroup: Rename to `additionalTripValidators`?

src/main/java/com/conveyal/gtfs/validator/NewTripTimesValidator.java
```diff
@@ -35,10 +35,10 @@
     private static final Logger LOG = LoggerFactory.getLogger(NewTripTimesValidator.class);
 
     int tripCount = 0;
+    boolean restrictValidators = false;
```
binh-dam-ibigroup: Maybe rename to `skipStandardTripValidation`?

src/main/java/com/conveyal/gtfs/validator/NewTripTimesValidator.java
```diff
@@ -47,13 +47,16 @@
 
     // As an optimization, these validators are fed the stoptimes for each trip to avoid repeated iteration and grouping.
     private final TripValidator[] tripValidators;
```
binh-dam-ibigroup: Maybe rename to `standardTripValidators`?

src/main/java/com/conveyal/gtfs/validator/FlexValidator.java
```diff
@@ -0,0 +1,495 @@
...
+        List<Location> locations
+    ) {
+        List<NewGTFSError> errors = new ArrayList<>();
+        if (locationGroupIdOrLocationIdIsStopId(stops, locationGroup.location_group_id)) {
+            errors.add(NewGTFSError.forEntity(
+                locationGroup,
+                NewGTFSErrorType.FLEX_FORBIDDEN_LOCATION_GROUP_ID).setBadValue(locationGroup.location_group_id)
+            );
+        }
+        if (locationGroupIdIsLocationId(locations, locationGroup.location_group_id)) {
```
binh-dam-ibigroup: Combine these two ifs.

binh-dam-ibigroup (approved):
Some minor suggestions, but I think we should move forward with this?

PR #355, opened 2022-07-15
Multiple Location Groups

br648:
### Checklist

### Description

This PR covers loading, exporting and validating multiple location groups. Multi location groups are defined as two or more location groups with the same id. These location groups on loading are combined into a single record/row with an array of location id values. On export, the reverse takes place expanding multi location groups in to single rows to conform with the GTFS Flex spec. A special case has been added to validate references where more than one location id has been defined i.e. it loops over all location ids and checks references for each.


miles-grant-ibigroup (approved):
Works great! Will merge into the "bigger" PR and then we can review further from there

PR #367, opened 2022-12-14
Refactor flex trip validation

br648:
### Checklist

### Description

Updated flex validation to validate only stop times that have flex columns.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/model/StopTime.java
+++ b/src/main/java/com/conveyal/gtfs/model/StopTime.java
@@ -212,7 +212,7 @@ public class StopTime extends Entity implements Cloneable, Serializable {
 
     /**
      * Check that the flex column 'start_pickup_dropoff_window' existing in the stop time table. If this column exists
-     * it is assumed that the other flex columns do to. This is to guard against cases where booking rules, location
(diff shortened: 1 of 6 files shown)


binh-dam-ibigroup (approved):
Seems to work and to pick up new valid errors on pickup/drop-off types that were not caught before.

src/main/java/com/conveyal/gtfs/model/StopTime.java
```diff
@@ -206,6 +210,76 @@ protected Iterator<StopTime> iterator() {
 
     }
 
+    /**
+     * Check that the flex column 'start_pickup_dropoff_window' existing in the stop time table. If this column exists
+     * it is assumed that the other flex columns do to. This is to guard against cases where booking rules, location
```
philip-cline:
```suggestion
     * it is assumed that the other flex columns do too. This is to guard against cases where booking rules, location
```

philip-cline (approved):

src/main/java/com/conveyal/gtfs/model/StopTime.java
br648: Addressed

PR #370, opened 2023-01-11
Flex trip pattern key update

br648:
### Checklist

### Description

This PR updates the trip pattern key equality check so that flex start/end pickup and drop off windows are also considered. This then creates additional patterns based on these values. Prior to this update trips were being grouped under the same, first, pattern which meant trips would have the wrong start/end pickup and drop off windows.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/TripPatternKey.java
+++ b/src/main/java/com/conveyal/gtfs/TripPatternKey.java
@@ -8,6 +8,7 @@ import gnu.trove.list.array.TIntArrayList;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import static com.conveyal.gtfs.model.Entity.INT_MISSING;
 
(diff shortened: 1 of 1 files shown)


philip-cline (approved):
Works well for me. This can help users a lot to differentiate patterns.

src/main/java/com/conveyal/gtfs/TripPatternKey.java
```diff
@@ -99,6 +102,8 @@ public int hashCode() {
         result = 31 * result + (stops != null ? stops.hashCode() : 0);
         result = 31 * result + (pickupTypes != null ? pickupTypes.hashCode() : 0);
         result = 31 * result + (dropoffTypes != null ? dropoffTypes.hashCode() : 0);
+        result = 31 * result + (start_pickup_dropoff_window != null ? start_pickup_dropoff_window.hashCode() : 0);
```
miles-grant-ibigroup: Does the 31 have to change now that we have added variables? Isn't there now a clash potential

miles-grant-ibigroup (approved):
Looks much cleaner! Just one quick question about the hashing

src/main/java/com/conveyal/gtfs/TripPatternKey.java
br648: Possibility. I've updated the method to use the newer approach to this:

`Objects.hash(routeId, stops, pickupTypes, dropoffTypes, start_pickup_dropoff_window, end_pickup_dropoff_window);`

PR #371, opened 2023-01-19
Refactor default travel time calculation

br648:
### Checklist

### Description

The process to calculate the previous departure time has been updated to include flex departures. If travelling from a point stop to a flex zone the point stop's departure time is used. If travelling from a flex zone to a point stop, the flex zone's end pickup drop off window is used. If travelling from a flex zone to another flex zone the previous departure time is set to zero. This is because a service will be making multiple trips between two zones and will therefore have many departure times.

Finally, the default travel time for a point stop is the arrival time minus the previous departure time. For a flex zone, the default travel time is the start pickup drop off window minus the previous departure time. However, if the previous stop was a flex zone, the default travel time is zero.


src/main/java/com/conveyal/gtfs/validator/PatternFinderValidator.java
```diff
@@ -301,6 +293,46 @@ public void complete(ValidationResult validationResult) {
...
+        Map<String, LocationGroup> locationGroupById,
+        String stopId) {
+        return locationById.containsKey(stopId) || locationGroupById.containsKey(stopId);
+    }
+
+    /**
+     * Calculate previous departure time, needed for all patterns. This is done by defining the 'last valid departure
+     * time' for all stops except the first.
+     */
+    private int calculatePreviousDepartureTime(
```
binh-dam-ibigroup: Could you add a unit test for this?

binh-dam-ibigroup (changes requested):
As mentioned, could you add a unit test for the travel time calculation?

src/main/java/com/conveyal/gtfs/validator/PatternFinderValidator.java
br648: Test added.

src/main/java/com/conveyal/gtfs/validator/PatternFinderValidator.java
```diff
@@ -161,64 +161,56 @@ public void complete(ValidationResult validationResult) {
                 TripPatternKey key = entry.getKey();
                 pattern.setStatementParameters(insertPatternStatement, true);
                 patternTracker.addBatch();
+                // Determine initial departure time based on the stop type.
+                int lastValidDepartureTime = isFlexStop(locationById, locationGroupById, key.stops.get(0))
+                    ? key.end_pickup_dropoff_window.get(0)
+                    : key.departureTimes.get(0);
```
binh-dam-ibigroup: This block is repeated almost the same at line 323 in this file and in the unit test at line 21, in both cases to fill-in the `stopSequence === 0` case. Could that code be moved to `calculatePreviousDepartureTime`?

binh-dam-ibigroup (commented):
@br648 Could you advise regarding https://github.com/conveyal/gtfs-lib/pull/371/files#r1094777561, please?

src/test/java/com/conveyal/gtfs/validator/CalculatePreviousDepartureTimeTest.java
```diff
@@ -39,96 +39,63 @@ void calculatePreviousDepartureTimeForTrip(TripPatternKey key, List<Departure> d
     private static Stream<Arguments> createTrips() {
         // Confirm that a trip that consists of just points (e.g. bus stops) will produce the correct departure times.
         TripPatternKey pointToPointKey = new TripPatternKey("test-route");
-        pointToPointKey.addStopTime(createStopTime(2, 0));
-        pointToPointKey.addStopTime(createStopTime(3, 0));
-        pointToPointKey.addStopTime(createStopTime(4, 0));
-        pointToPointKey.addStopTime(createStopTime(5, 0));
-        pointToPointKey.addStopTime(createStopTime(6, 0));
+        pointToPointKey.addStopTime(createStopTime("1",2, 0));
```
binh-dam-ibigroup: Insert space after `"1",`.

binh-dam-ibigroup: Nit: I would have prefixed the numbers in quotes with `stop-id-` to clarify.

src/main/java/com/conveyal/gtfs/validator/PatternFinderValidator.java
```diff
@@ -304,17 +297,24 @@ private boolean isFlexStop(
     }
 
     /**
-     * Calculate previous departure time, needed for all patterns. This is done by defining the 'last valid departure
+     * Calculate previous departure times, needed for all patterns. This is done by defining the 'last valid departure
      * time' for all stops except the first.
```
binh-dam-ibigroup: Insert a comment about the assumption of what the previous departure time is for `stopSequence = 0`, and maybe repeat that comment in the test where the consecutive `2, 2` departure time sequence appears.

binh-dam-ibigroup (approved):
The code needs one comment regarding an assumption made. (Changing that assumption is out of scope for the PR).

src/main/java/com/conveyal/gtfs/validator/PatternFinderValidator.java
br648: I've updated to always return zero for the first stop which is much clearer.

philip-cline (approved):
Looks good to me.

PR #377, opened 2023-04-18
Changed the prior notice start/last time to String

br648:
### Checklist

### Description

Changed the prior notice start and last time from time fields (int) to string fields. This is to allow the UI to enter times in a more user friendly format e.g. hh:mm:ss instead of seconds from midnight. Validation is not effected because only the presence of the start time field is checked (no the actual value).


philip-cline (approved):
Looks good to me and I ran some tests on it which worked well!

miles-grant-ibigroup (approved):
Fantastic! Thanks for this

PR #385, opened 2023-06-22
Exception based scheduling

br648:
### Checklist

### Description

The crux of this feature is to allow calendar dates (which are not associated to a calendar, as this is already imlemented) to be loaded, snapshotted and exported without the need to be joined to a calendar. This then allows calendar date services to be defined in their own right and not tied to a calendar.  

From the spec (https://developers.google.com/transit/gtfs/reference#calendar_datestxt) this will allow the alternative approach to using calendar dates.

It should be possible under the UI to define a calendar date service by providing the following:

- Exception name. This will be used as the service id, so will need to be unique.
- New schedule type of "CALENDAR_DATE_SERVICE" with a value of 10.
- A single date.

It is assumed that:

- Only active service dates are provided (exception_type: 1).
- That the existing process for submitting this data to the backend will not need altering (testing will confirm either way).


src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -126,38 +128,57 @@ public FeedLoadResult exportTables() {
...
                     // feed does have calendars.txt file, continue export with strategy of matching exceptions
                     // to calendar to output calendar_dates.txt
-                    int calendarDateCount = 0;
                     for (Calendar cal : calendars) {
                         Service service = new Service(cal.service_id);
                         service.calendar = cal;
-                        for (ScheduleException ex : exceptions.stream()
+                        for (ScheduleException ex : calendarExceptions.stream()
                             .filter(ex -> exceptionInvolvesService(ex, cal.service_id))
                             .collect(Collectors.toList())
```
binh-dam-ibigroup, line 185: Replace `for... collect(...)` with just `.forEach`?

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -126,38 +128,57 @@ public FeedLoadResult exportTables() {
...
+                    CalendarDate calendarDate = new CalendarDate();
+                    calendarDate.date = date;
+                    calendarDate.service_id = ex.name;
+                    calendarDate.exception_type = 1;
+                    Service service = new Service(calendarDate.service_id);
+                    service.calendar_dates.put(date, calendarDate);
+                    // If the calendar dates provided contain duplicates (e.g. two or more identical service ids that are
+                    // NOT associated with a calendar) only the first entry will persist export.
+                    feed.services.put(calendarDate.service_id, service);
+                    calendarDateCount++;
```
binh-dam-ibigroup: Are we expecting errors? Otherwise, `calendarDateCount` should be the same as `calendarDateExceptions.size()`.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsSnapshotter.java
```diff
@@ -276,20 +299,23 @@ private TableLoadResult createScheduleExceptionsTable() {
...
-                    Table.CALENDAR,
-                    dataSource,
-                    feedIdToSnapshot + ".",
-                    EntityPopulator.CALENDAR
-                );
-                Set<String> calendarServiceIds = new HashSet<>();
-                for (Calendar calendar : calendarReader.getAll()) {
-                    calendarServiceIds.add(calendar.service_id);
+                for (Map.Entry<String,String> entry : calendarDateService.entrySet()) {
+                    String serviceId = entry.getValue();
```
binh-dam-ibigroup: Extract a method for the similar content of this for loop and the one above at line 287.

src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
```diff
@@ -1442,15 +1467,19 @@ private static Set<Table> getReferencingTables(Table table) {
...
+                            referencingTables.add(gtfsTable);
+                        }
+                    }
                 }
             }
         }
         return referencingTables;
     }
 
+
```
binh-dam-ibigroup: Not needed.

src/main/java/com/conveyal/gtfs/loader/ReferenceTracker.java
```diff
@@ -79,15 +82,30 @@ public Set<NewGTFSError> checkReferencesAndUniqueness(String keyValue, int lineN
...
+            // Note: The reference table must be loaded before the table/value being currently checked.
+            boolean hasMatchingReference = false;
+            TreeSet<String> badValues = new TreeSet<>();
+            for (Table referenceTable : field.referenceTables) {
+                String referenceField = referenceTable.getKeyFieldName();
+                if (checkReference(referenceField, value, badValues)) {
+                    hasMatchingReference = true;
+                    break;
+                }
+            }
```
binh-dam-ibigroup: Extract method (move comments at lines 85-87 into method), inline result into if condition at line 97, delete comment at line 98.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsSnapshotter.java
```diff
@@ -231,41 +236,59 @@ private TableLoadResult createScheduleExceptionsTable() {
...
+                            // copying from a feed that doesn't have the calendar.txt file
+                            Calendar calendar = dummyCalendarsByServiceId.getOrDefault(calendarDate.service_id, new Calendar());
+                            calendar.service_id = calendarDate.service_id;
+                            if (calendar.start_date == null || calendar.start_date.isAfter(calendarDate.date)) {
+                                calendar.start_date = calendarDate.date;
+                            }
+                            if (calendar.end_date == null || calendar.end_date.isBefore(calendarDate.date)) {
+                                calendar.end_date = calendarDate.date;
+                            }
+                            dummyCalendarsByServiceId.put(calendarDate.service_id, calendar);
```
binh-dam-ibigroup: Probably extract a method.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsSnapshotter.java
```diff
@@ -210,15 +211,19 @@ private TableLoadResult createScheduleExceptionsTable() {
...
+                JDBCTableReader<Calendar> calendarReader = new JDBCTableReader(
+                    Table.CALENDAR,
+                    dataSource,
+                    feedIdToSnapshot + ".",
+                    EntityPopulator.CALENDAR
                 );
+                Set<String> calendarServiceIds = new HashSet<>();
+                for (Calendar calendar : calendarReader.getAll()) {
+                    calendarServiceIds.add(calendar.service_id);
+                }
```
binh-dam-ibigroup: Extract method that is common with code from `JdbcGtfsExporter`, lines 133-140?

binh-dam-ibigroup (commented):
Need to play with this code more, but see my comments so far.

binh-dam-ibigroup (approved):
I'm able to replicate a correct export of a schedule exception not linked to a calendar (with some cheating), so I will go ahead and approve subject to the refactorings I mentioned in my previous comment.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
br648: "Variable used in lambda expression should be final or effectively final". So leaving as is.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -126,38 +128,54 @@ public FeedLoadResult exportTables() {
...
+                for (ScheduleException ex : exceptionsIterator) {
+                    if (ex.exemplar.equals(ScheduleException.ExemplarServiceDescriptor.CALENDAR_DATE_SERVICE)) {
+                        calendarDateExceptions.add(ex);
+                    } else {
+                        calendarExceptions.add(ex);
+                    }
                 }
+
+                int calendarDateCount = calendarDateExceptions.size();
+                // Extract calendar date services, convert to calendar date and add to the feed. We are expect only one
```
philip-cline:
```suggestion
                // Extract calendar date services, convert to calendar date and add to the feed. We are expecting only one
```

src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
```diff
@@ -1563,71 +1591,75 @@ private void updateReferencingTables(
...
-                                        // Append referenced patterns data to the end of the error.
-                                        message = String.format(
-                                            "%s\nReferenced patterns: [%s]",
-                                                message,
-                                                StringUtils.join(patternAndRouteIds, ",")
-                                        );
-                                    }
-                                    LOG.warn(message);
-                                    throw new SQLException(message);
+                                    LOG.info("{} reference(s) in {} {}D!", result, refTableName, sqlMethod);
```
philip-cline, line 1717: Is the `D` intentional? 
```suggestion
                                    LOG.info("{} reference(s) in {} {}!", result, refTableName, sqlMethod);
```

philip-cline (changes requested):
Code looks good, but right now I'm a bit confused about the use of the `ScheduleException` name as the service_id for a calendar_date based service. For example, calendar_dates may be used to add service on 2023-05-29 and 2023-07-04 for a service ID "4" which does not exist in calendar.txt. These would be represented as two separate calendar_dates but right now result in two separate ScheduleExceptions but with the same `name`. How should this use case be handled if we requiring name to be unique?
image

src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java
br648: Yep! The SQL method is either UPDATE or DELETE and the 'D' makes it past tense.

br648:
> Code looks good, but right now I'm a bit confused about the use of the `ScheduleException` name as the service_id for a calendar_date based service. For example, calendar_dates may be used to add service on 2023-05-29 and 2023-07-04 for a service ID "4" which does not exist in calendar.txt. These would be represented as two separate calendar_dates but right now result in two separate ScheduleExceptions but with the same `name`. How should this use case be handled if we requiring name to be unique? image

@philip-cline How does this look? If the calendar date is not assoicated with a calendar the name is made up of service id and the first date. For `only-in-calendar-dates-txt-20170916` there is only one entry so strictly speaking the appended date is not needed, but for consistency I've kept it.

image

philip-cline:
Thanks yeah this is looking good. I'll start on a front end PR for this.

philip-cline (approved):
LGTM to me now

PR #387, opened 2023-07-06
Allow exception overlap

philip-cline:
### Checklist

### Description

This PR modifies the restrictions on exception dates to allow overlap. 

Companion PR, with context of the fix: https://github.com/ibi-group/datatools-ui/pull/965

### DIFF
--- a/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
+++ b/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
@@ -159,7 +159,7 @@ public class JdbcGtfsExporter {
                                 calendarDate.date = date;
                                 calendarDate.service_id = cal.service_id;
                                 calendarDate.exception_type = ex.serviceRunsOn(cal) ? 1 : 2;
-                                LOG.info("Adding exception {} (type={}) for calendar {} on date {}", ex.name, calendarDate.exception_type, cal.service_id, date.toString());
+                                LOG.info("Adding exception {} (type={}) for calendar {} on date {}", ex.name, calendarDate.exception_type, cal.service_id, date);
 
                                 if (service.calendar_dates.containsKey(date))
                                     throw new IllegalArgumentException("Duplicate schedule exceptions on " + date.toString());
--- a/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
+++ b/src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
@@ -69,6 +69,17 @@ public class JdbcGtfsExporter {
     }
 
     /**
+     * Utility method to check if an exception uses a specific service.
+     */
+    public Boolean exceptionInvolvesService(ScheduleException ex, String serviceId) {
+        return (
+            ex.addedService.contains(serviceId) ||
+            ex.removedService.contains(serviceId) ||
+            ex.customSchedule.contains(serviceId)
+        );
+    }
+
+    /**
      * Export primary entity tables as well as Pattern and PatternStops tables.
      *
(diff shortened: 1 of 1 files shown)


br648 (approved):
LGTM.

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -153,7 +159,7 @@ public FeedLoadResult exportTables() {
                                 calendarDate.date = date;
                                 calendarDate.service_id = cal.service_id;
                                 calendarDate.exception_type = ex.serviceRunsOn(cal) ? 1 : 2;
-                                LOG.info("Adding exception {} (type={}) for calendar {} on date {}", ex.name, calendarDate.exception_type, cal.service_id, date.toString());
+                                LOG.info("Adding exception {} (type={}) for calendar {} on date {}", ex.name, calendarDate.exception_type, cal.service_id, date);
```
miles-grant-ibigroup, line 171: Is date now a string?

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
```diff
@@ -135,7 +136,12 @@ public FeedLoadResult exportTables() {
                     for (Calendar cal : calendars) {
                         Service service = new Service(cal.service_id);
                         service.calendar = cal;
-                        for (ScheduleException ex : exceptions) {
+                        List<ScheduleException> calendarExceptions = exceptions.stream().filter(ex ->
+                            ex.addedService.contains(cal.service_id) ||
+                            ex.removedService.contains(cal.service_id) ||
+                            ex.customSchedule.contains(cal.service_id)
+                        ).collect(Collectors.toList());
+                        for (ScheduleException ex : calendarExceptions) {
```
miles-grant-ibigroup: You can probably use a `.forEach` on this stream to avoid the collector or at least inline this!

miles-grant-ibigroup (changes requested):
Works great, just some small cleanup

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
philip-cline: using `toString` is just redundant according to SonarLint. 
image

src/main/java/com/conveyal/gtfs/loader/JdbcGtfsExporter.java
philip-cline: We have a counter that we're using inside of the for loop which means that this will get complicated if we try to use a lambda function. In light of this, I think this should stay as is.

miles-grant-ibigroup: Makes sense

miles-grant-ibigroup: Can we put the stream inside the for statement instead of assigning to a variable?

philip-cline: Sure, that's done in https://github.com/conveyal/gtfs-lib/pull/387/commits/e51456e60c8761f646a7bcb71e600676c49606aa although it makes the code maybe a bit harder to read

miles-grant-ibigroup: That's not what I meant I meant instead of saving the collected stream to a variable, put the entire stream inside the for definition

philip-cline: See inline filtering now please

miles-grant-ibigroup (approved):
Looks great! Thanks for the changes

PR #390, opened 2023-07-25
fix: add support for externally defined `newGTFSErrorType`

miles-grant-ibigroup:
This is facilitated by having `newGTFSErrorType` implement a new interface. This interface can be implemented by external code as well, allowing the internal error methods to make use of error types gtfs-lib does not know about.


miles-grant-ibigroup:
This can maybe be useful in the future, hopefully

PR #394, opened 2023-08-24
Persist patterns

br648:
### Checklist

### Description

Refactor to persist pattern data on export and load on import. The PR also refectors pattern building by creating a new PatternBuilder class for this process with distinct methods instead of a series of tasks in PatternFinderValidator.

### DIFF
--- a/src/main/java/com/conveyal/gtfs/PatternBuilder.java
+++ b/src/main/java/com/conveyal/gtfs/PatternBuilder.java
@@ -40,14 +40,10 @@ public class PatternBuilder {
         connection = feed.getConnection();
     }
 
-    private String getTableName(String tableName) {
-        return feed.tablePrefix + tableName;
-    }
-
     public void create(Map<TripPatternKey, Pattern> patterns, Set<String> patternIdsLoadedFromFile) {
-        String patternsTableName = getTableName("patterns");
-        String tripsTableName = getTableName("trips");
-        String patternStopsTableName = getTableName("pattern_stops");
+        String patternsTableName = feed.getTableName("patterns");
+        String tripsTableName = feed.getTableName("trips");
+        String patternStopsTableName = feed.getTableName("pattern_stops");
 
         Table patternsTable = new Table(patternsTableName, Pattern.class, Requirement.EDITOR, Table.PATTERNS.fields);
         Table patternStopsTable = new Table(patternStopsTableName, PatternStop.class, Requirement.EDITOR, Table.PATTERN_STOP.fields);
@@ -62,9 +58,10 @@ public class PatternBuilder {
                 patternsTable.createSqlTable(connection, null, true);
             }
             patternStopsTable.createSqlTable(connection, null, true);
-            PrintStream patternForTripsFileStream = createTempPatternForTripsTable(tempPatternForTripsTextFile, statement);
-            processPatternAndPatternStops(patternsTable, patternStopsTable, patternForTripsFileStream, patterns, patternIdsLoadedFromFile);
-            updateTripPatternIds(tempPatternForTripsTextFile, patternForTripsFileStream, statement, tripsTableName);
+            try (PrintStream patternForTripsFileStream = createTempPatternForTripsTable(tempPatternForTripsTextFile, statement)) {
+                processPatternAndPatternStops(patternsTable, patternStopsTable, patternForTripsFileStream, patterns, patternIdsLoadedFromFile);
+            }
+            updateTripPatternIds(tempPatternForTripsTextFile, statement, tripsTableName);
             createIndexes(statement, patternsTableName, patternStopsTableName, tripsTableName);
             connection.commit();
         } catch (SQLException | IOException e) {
@@ -143,13 +140,11 @@ public class PatternBuilder {
      */
     private void updateTripPatternIds(
         File tempPatternForTripsTextFile,
-        PrintStream patternForTripsFileStream,
         Statement statement,
         String tripsTableName
     ) throws SQLException, IOException {
 
         LOG.info("Updating trips with pattern IDs.");
-        patternForTripsFileStream.close();
         // Copy file contents into temp pattern for trips table.
         copyFromFile(connection, tempPatternForTripsTextFile, TEMP_FILE_NAME);
         // Before updating the trips with pattern IDs, index the table on trip_id.
--- a/src/main/java/com/conveyal/gtfs/PatternBuilder.java
+++ b/src/main/java/com/conveyal/gtfs/PatternBuilder.java
@@ -143,7 +143,6 @@ public class PatternBuilder {
         Statement statement,
         String tripsTableName
     ) throws SQLException, IOException {
-
         LOG.info("Updating trips with pattern IDs.");
         // Copy file contents into temp pattern for trips table.
         copyFromFile(connection, tempPatternForTripsTextFile, TEMP_FILE_NAME);
@@ -190,7 +189,6 @@ public class PatternBuilder {
      * Construct pattern stops based on values in trip pattern key.
      */
     private void createPatternStops(TripPatternKey key, String patternId, Table patternStopsTable) throws SQLException {
-
         String insertPatternStopSql = patternStopsTable.generateInsertSql(true);
         PreparedStatement insertPatternStopStatement = connection.prepareStatement(insertPatternStopSql);
         BatchTracker patternStopTracker = new BatchTracker("pattern stop", insertPatternStopStatement);
--- a/src/main/java/com/conveyal/gtfs/PatternBuilder.java
+++ b/src/main/java/com/conveyal/gtfs/PatternBuilder.java
@@ -1,0 +1,242 @@
+package com.conveyal.gtfs;
+
+import com.conveyal.gtfs.loader.BatchTracker;
+import com.conveyal.gtfs.loader.Feed;
+import com.conveyal.gtfs.loader.Requirement;
+import com.conveyal.gtfs.loader.Table;
+import com.conveyal.gtfs.model.Pattern;
+import com.conveyal.gtfs.model.PatternStop;
+import org.apache.commons.dbutils.DbUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.PrintStream;
+import java.nio.file.Files;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Map;
+import java.util.Set;
+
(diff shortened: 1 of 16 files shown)


src/test/java/com/conveyal/gtfs/GTFSTest.java
```diff
@@ -996,6 +997,15 @@ private static int countValidationErrorsOfType(
...
 
+    /**
+     * Proprietary table file names are prefix with "datatools_" to distinguish them from GTFS spec files.
+     */
+    private String getTableFileName(String tableName) {
+        return (tableName.equals("patterns"))
+            ? String.format("datatools_%s.txt", tableName)
+            : tableName + ".txt";
+    }
+
```
binh-dam-ibigroup: This method duplicates `getFileName` in `src/main/java/com/conveyal/gtfs/loader/Table.java`. Please merge the two and name the resulting method `getTableFileName`.

src/main/java/com/conveyal/gtfs/PatternBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+
+    private final Connection connection;
+    public PatternBuilder(Feed feed) throws SQLException {
+        this.feed = feed;
+        connection = feed.getConnection();
+    }
+
+    private String getTableName(String tableName) {
+        return feed.tablePrefix + tableName;
+    }
```
binh-dam-ibigroup: Maybe move to a utils class. This construct is used many times in the codebase.

src/main/java/com/conveyal/gtfs/PatternBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+     */
+    private void updateTripPatternIds(
+        File tempPatternForTripsTextFile,
+        PrintStream patternForTripsFileStream,
+        Statement statement,
+        String tripsTableName
+    ) throws SQLException, IOException {
+
+        LOG.info("Updating trips with pattern IDs.");
+        patternForTripsFileStream.close();
```
binh-dam-ibigroup: Why is `patternForTripsFileStream` passed to this method? The only action on it is to close it... How about using a try-with-resources in the method that calls this one?

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -487,6 +495,15 @@ public Table addPrimaryKeyNames(String ...keys) {
         return this;
     }
 
+    /**
+     * Fluent method that defines tables that on export should have the proprietary prefixed value applied to the file
+     * name.
+     */
+    public Table hasProprietaryFilePrefix() {
```
binh-dam-ibigroup: Rename this method to `isProprietary` or `makeProprietary`?

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -103,6 +103,12 @@ public class Table {
     /** Key(s) that uniquely identify an entry in the respective table.*/
     private String[] primaryKeyNames;
 
+    /** Prefixed to exported tables/files that are not part of the GTFS spec. */
+    public static String proprietaryFilePrefix = "datatools_";
```
binh-dam-ibigroup: Make final. Use uppercase naming?

binh-dam-ibigroup (commented):
Here are my first comments (I need to play some more):

src/main/java/com/conveyal/gtfs/PatternBuilder.java
br648: I've moved it into the Feed class as it is related to getting the database table name with schema prefix as opposed to just table name.

src/main/java/com/conveyal/gtfs/loader/Table.java
br648: Now that the method `getTableFileName` is static, this method is no longer needed.

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -635,11 +622,13 @@ public String generateInsertSql (String namespace, boolean setDefaultId) {
...
-        return this.isProprietaryTable
-            ? Table.proprietaryFilePrefix + fileName
-            : fileName;
+    /**
+     * Proprietary table file names are prefix with "datatools_" to distinguish them from GTFS spec files.
+     */
+    public static String getTableFileName(String tableName) {
+        return (tableName.equals("patterns"))
+            ? String.format("%s%s.txt", PROPRIETARY_FILE_PREFIX, tableName)
+            : tableName + ".txt";
```
binh-dam-ibigroup: Can we use a formatted string for the non-proprietary case too?

src/main/java/com/conveyal/gtfs/model/Pattern.java
```diff
@@ -114,7 +113,7 @@ public Pattern () {}
     public static class Loader extends Entity.Loader<Pattern> {
 
         public Loader(GTFSFeed feed) {
-            super(feed, Table.proprietaryFilePrefix + "patterns");
+            super(feed, Table.PROPRIETARY_FILE_PREFIX + "patterns");
```
binh-dam-ibigroup: This and the ctor at line 144, and also the code from `getTableFileName`: extract logic to build the table file name without the extension. Maybe add an overload to `getTableName` to accept the table prefix?

binh-dam-ibigroup (approved):
Works well! See a few tweaks.

src/main/java/com/conveyal/gtfs/validator/PatternFinderValidator.java
```diff
@@ -66,172 +72,18 @@ public void validateTrip (Trip trip, Route route, List<StopTime> stopTimes, List
      */
     @Override
     public void complete(ValidationResult validationResult) {
+        Set<String> patternIds = new HashSet<>();
+        for(Pattern pattern :  feed.patterns) {
+            patternIds.add(pattern.pattern_id);
```
philip-cline, line 77: Can we make use of the `TableReader#getAll` method here rather than a for loop?

src/main/java/com/conveyal/gtfs/PatternBuilder.java
```diff
@@ -0,0 +1,237 @@
...
+    /**
+     * Copy the pattern for trips text file into a table, create an index on trip IDs, and update the trips
+     * table.
+     */
+    private void updateTripPatternIds(
+        File tempPatternForTripsTextFile,
+        Statement statement,
+        String tripsTableName
+    ) throws SQLException, IOException {
+
```
philip-cline: nit: maybe remove this empty line

src/main/java/com/conveyal/gtfs/PatternBuilder.java
```diff
@@ -0,0 +1,237 @@
...
+        // Before updating the trips with pattern IDs, index the table on trip_id.
+        String patternForTripsIndexSql = String.format(
+            "create index temp_trips_pattern_id_idx on %s (trip_id)",
+            TEMP_FILE_NAME
+        );
+        LOG.info(patternForTripsIndexSql);
+        statement.execute(patternForTripsIndexSql);
+        // Finally, execute the update statement.
+        String updateTripsSql = String.format(
+            "update %s set pattern_id = %s.pattern_id from %s where %s.trip_id = %s.trip_id",
```
philip-cline, line 158: Would including a condition here or elsewhere where only pattern IDs which have changed are updated have any performance benefit?

philip-cline: The trips table is usually not very big so I'm not sure this is very important

src/main/java/com/conveyal/gtfs/loader/Table.java
```diff
@@ -103,6 +103,9 @@ public class Table {
     /** Key(s) that uniquely identify an entry in the respective table.*/
     private String[] primaryKeyNames;
 
+    /** Prefixed to exported tables/files that are not part of the GTFS spec. */
+    public static final String PROPRIETARY_FILE_PREFIX = "datatools_";
```
philip-cline, line 107: I'm not completely sure I love the `datatools_` prefix. Maybe something more declarative would be better? Like `namemapping_patterns.txt` or something? Thoughts?

src/main/java/com/conveyal/gtfs/PatternBuilder.java
```diff
@@ -0,0 +1,237 @@
...
+        String patternStopsTableName = feed.getTableNameWithSchemaPrefix("pattern_stops");
+
+        Table patternsTable = new Table(patternsTableName, Pattern.class, Requirement.EDITOR, Table.PATTERNS.fields);
+        Table patternStopsTable = new Table(patternStopsTableName, PatternStop.class, Requirement.EDITOR, Table.PATTERN_STOP.fields);
+
+        try {
+            File tempPatternForTripsTextFile = File.createTempFile(TEMP_FILE_NAME, "text");
+            LOG.info("Creating pattern and pattern stops tables.");
+            Statement statement = connection.createStatement();
+            statement.execute(String.format("alter table %s add column pattern_id varchar", tripsTableName));
```
philip-cline, line 55: This feels a little dangerous maybe to alter the table to include the pattern_id here on every run? Is this because the old structure did not include the `pattern_id`?

philip-cline (changes requested):
Looking rly good, just a couple of questions

src/main/java/com/conveyal/gtfs/validator/PatternFinderValidator.java
br648: Unfortunately the getAll method only returns an iterator so I would still have to loop through each item to get the pattern id.

src/main/java/com/conveyal/gtfs/PatternBuilder.java
br648: I think the overhead of managing that would outweigh just making an update.

src/main/java/com/conveyal/gtfs/PatternBuilder.java
br648: It's because the GTFS structure for trips doesn't include patterns which is a DT thing. The data is loaded in under the context of GTFS and then altered to work with patterns.

src/main/java/com/conveyal/gtfs/loader/Table.java
br648: `datatools_` is sad that you are not showing the love! 

It's the complete pattern data not just the name. I like/love `datatools_` because it defines the origin.

philip-cline: If you like it, I don't feel that strongly about it! It definitely identifies it well as a custom file that's coming out of datatools

src/main/java/com/conveyal/gtfs/PatternBuilder.java
philip-cline: Gotcha, that seems fine

philip-cline (approved):
This is looking good, thanks again!

